PR #75, opened 2015-11-20
updated product URLs

mah11:
use GitHub raw links; added OBO file URL


jamesaoverton:
These test URLs should be updated in about 20 minutes: 

- http://url.ontodev.org/obo/fypo.owl
- http://url.ontodev.org/obo/fypo.obo

The PURLs will be updated when the new system is deployed, which is planned for Monday.

PR #76, opened 2015-11-20
Github tracker

drseb:
We have moved the tracker from sourceforge to github.


jamesaoverton:
This test URL should be updated in about 15 minutes: http://url.ontodev.org/obo/hp/tracker

The PURL will be updated when this new system is deployed, which is planned for Monday.

Issue #64, opened 2015-11-18
PURLs for APOLLO_SV


jamesaoverton:
I don't see this project in the registry, but PURL.org has entries for it. What to do? @alanruttenberg

jamesaoverton:
Current term PURLs do not resolve properly. Compare: 

- http://purl.obolibrary.org/obo/APOLLO_SV_00000219
- http://www.ontobee.org/ontology/APOLLO_SV?iri=http://purl.obolibrary.org/obo/APOLLO_SV_00000219

I think it's because the `/obo/apollo_sv/about/` entry is misconfigured on PURL.org.

alanruttenberg:
Yes, the about entry is incorrect. It sets the ?iri= to http://purl.obolibrary.org/apollo_sv/ instead of http://purl.obolibrary.org/obo/

<pre>
id: /obo/apollo_sv/about/
type: partial
target: http://www.ontobee.org/browser/rdf.php?o=APOLLO_SV&iri=http://purl.obolibrary.org/apollo_sv/
maintainers: ALANRUTTENBERG,hoganwr,mbrochhausen,apollo_sv
status: Approved
</pre>

The entry for apollo_sv is in http://code.google.com/p/obo-registry/source/browse/trunk/metadata/ontologies.txt  

I'll create an entry for ontologies/ and commit it.

alanruttenberg:
https://github.com/OBOFoundry/purl.obolibrary.org/pull/77

Please review before committing - first time creating one of these- and then close.

jamesaoverton:
Closed by commit 116a98b40c607d543051679c3730e41d37351934

PR #77, opened 2015-11-21
Create apollo_sv.yaml

alanruttenberg:
PURL configuration for http://purl.obolibrary.org/obo/apollo_sv/. Note this is a mixture of pointers to releases on googlecode and development on github. Should be fixed..


jamesaoverton:
Closes #64

PR #92, opened 2015-11-23
Add tests for case insensitive matching

jamesaoverton:
- these actually test `config/chebi.yml`,
  but our scoping rules require the tests be in `config/obi.yml`


jamesaoverton:
Oops. I don't want to merge this yet.

Issue #97, opened 2015-11-24
Many unregistered IDSPACEs are in use


jamesaoverton:
In the past 24 hours I've seen requests for PURLs in these IDSPACEs that are not in the [OBO registry](https://github.com/OBOFoundry/OBOFoundry.github.io/tree/master/ontology). Some I recognize, but most I do not. (I compiled this list manually, so there may be some mistakes.)

- ABA
- ARG
- AcarDv
- B8467
- BILS
- CDT
- CGNC
- COLLECTION
- CS
- DERMO
- DHBA
- DrerDv
- EFO
- EKB
- ENSEMBL
- EXAMPLE
- EcoGene
- FB
- FBql
- FF
- FlyBase
- GENO
- GOCHE
- GgalDv
- HGNC
- HINO
- ID
- INO
- MBA
- MCBCC
- MGI
- MOBO
- MdomDv
- NCBIGene
- OARCS
- OBIws
- ODNAE
- OHD
- OMIABIS
- OMIM
- Orphanet
- PANTHER
- PBA
- PDO
- PHI
- PRO
- Pfam
- PomBase
- REO
- RGD
- SGD
- SIO
- SMTH
- TAIR
- UBPROP
- UCHEBI
- UniProtKB
- VICO
- XEO
- ZFIN
- ZP

alanruttenberg:
A known problem that we're finally in a position to fix. We should go through these trying to find their owners and ensure they are registered properly. For ones that we can't identify we should write to obo-discuss with the list and warning that unless registered we will start to not let them through in some reasonable amount of time.

Of the ones there, I know
PRO - Not sure what to do - probably should find out who is using this and inform them to change to "PR_"
Pfam - I wonder if PRO is using these
OHD - Mine - thought I registered but I guess not. I will register it. Oral Health and Disease Ontology
CDT - Mine used in OHD. I will register.
OMIABIS - Contact Mathias Brochhausen mbrochhausen@users.sf.net
OBIws - Contact Jie Zheng
REO - Contact Melissa Haendel
SIO - Contact Michel Dumontier

There are others that are familiar but I'm not sure their status. Would it be possible to get a sampling of the domains that are requesting them, to see which are used generally and which might be used by only a small number of hosts?

I suspect a number of them are autogenerated from dbxrefs. 

Presumably these are falling through to berkeleybop?

Personally I'd like to get away from any automatic fallthrough like the one we are using now.

[...]

PR #111, opened 2015-11-30
cl fall-through

cmungall:
See #97


jamesaoverton:
Should be fine now. The problem last time was a conflict between the tests automatically generated for two rules `exact: /` and `prefix: /`.

Looks like the branch has to be update first, though.

cmungall:
I previously tagged this with #97 but I now see that the two are unconnected. The spreadsheet in 97 has URLs like /obo/cl/imports/chebi_[DIGITS]import.[DIGITS]owl which are obviously bogus and the result of some script barfing

PR #115, opened 2015-12-01
Fix typo: move xap.yml to xao.yml

jamesaoverton:
Looks like I made a typo during the original migration: 'xap.yml'.

@cmungall Please confirm the new config is correct.


cmungall:
The tracker is out of date but we can fix later

In retrospect a purl for the tracker is OTT and not required, and no one uses them

Issue #121, opened 2015-12-07
Something odd is happening with a number of OWL served via PURLs


cmungall:
May be related to #109

@drseb reported this

`owltools http://purl.obolibrary.org/obo/hp.owl`

fails. This is replicated in Protege 5. Downloading the file and parsing it works fine

@kltm reported the same thing for http://purl.obolibrary.org/obo/mp.owl

jamesaoverton:
Downloading `hp.owl` from the PURL worked for me. Using the PURL in Protege 5 failed once and succeeded once.

cmungall:
I think it's always been an issue with mp now I think about it, FTP issue

hp is a mystery.

hdietze:
I still have issues loading the owl file via the OWL-API (3.5.1), it seems the file is incomplete and the RDF parser errors around line numbers like 121578, 115218, or 115620. This seems to indicate a buffering issue in the OWL-API Java client. Using command-line wget seems to work fine.

PR #122, opened 2015-12-08
Moved from hudson to jenkins

drseb:


cmungall:
@hdietze can you comment re forward to https. Any issues here?

hdietze:
The https seems to work fine. Doesn't seem to change #121 though.

PR #150, opened 2016-02-05
Create MFOMD.yml to appropriately redirect

jannahastings:

PR #151, opened 2016-02-05
Add mfomd.yml with HTTPS links

jamesaoverton:


jamesaoverton:
This is a minor change to #150, which I failed to update correctly.

PR #154, opened 2016-02-05
Unquote URLs everywhere

jamesaoverton:


jamesaoverton:
This solves a problem with #153, it passes all the tests, and I believe that it's correct. I'm going to merge it, but I'm still a little nervous. @cmungall: Please keep an eye out for percent-encoding problems that might be caused by this patch.

PR #161, opened 2016-02-17
Update iao.yml

mbrochhausen:
added a new PURL for a new version of d-acts.


cmungall:
looks good. But I recommend using the github release mechanism rather than adding versioned artefacts into master

mbrochhausen:
Chris,

thanks. Yes, I plan to use tags (I hope that is what you are referring to)
for releases. However, this is not a release, but a (potentially)
controversial revision of the last release version. Optimally, I would have
done that as a new branch, but I was working under considerable time
constraint and needed that version up and accessible for participants at a
workshop.

PR #176, opened 2016-04-22
Update ovae.yml

yongqunh:
I updated ovae.yml. One main change was the update of the owl file pointed by the PURL.
Oliver


yongqunh:
Thanks!

PR #177, opened 2016-04-22
Update iao.yml

mbrochhausen:
redirected d-acts.owl and dev/d-acts.owl


jamesaoverton:
I just merged Pull Request #176 into master, so you'll have to update this PR before it can be merged. There should be an obvious way to do that, but if it's not clear, just ask me.

mbrochhausen:
Done (I think)

Issue #106, opened 2015-11-26
Simpler YAML configuration


jamesaoverton:
I've created the [simpler](https://github.com/OBOFoundry/purl.obolibrary.org/tree/simpler) branch to experiment with a simpler and more consistent YAML configuration format. It addresses most of the complaints I made in https://github.com/OBOFoundry/purl.obolibrary.org/issues/100#issuecomment-159391144, but is really orthogonal to #100.

The old format omitted the `base_url` (e.g. `/obo/obi`) from most paths. This was designed to save a little typing and enforce separation of project-level rules from top-level rules. But each project has *some* top-level rules, and to handle them I had to create `products`, `term_browser`, and `base_redirect` fields, each with different behaviour.

In the new format the `base_url` is always included. The `translate.py` script operates in two "modes": project and top. In project mode it only generates Apache directives in the project's space, e.g. under `/obo/obi/`. In top mode it only generates Apache directives in the top-level `/obo/` space (not project space), and restricts them to the given IDSPACE so that only these patterns work:

- `/obo/obi`
- `/obo/obi.owl`
- `/obo/obi.obo`
- term browser: ontobee
- top-level regex rules: we should have very few of these, but we have to check them manually

Full examples: [OBI.yml](https://github.com/OBOFoundry/purl.obolibrary.org/blob/simpler/config/OBI.yml), [CHEBI](https://github.com/OBOFoundry/purl.obolibrary.org/blob/simpler/config/CHEBI.yml)

The upshot is less code with fewer special cases. I've also tweaked the nomenclature for clarity and updated the documentation.

alanruttenberg:
So do I understand that 'browser: ontobee' will automatically add the /obo/<namespace>/about/ redirect?

I suggest we also add an alias(or maybe use YAML feature if I understand it) and a  release directive for a start

e.g. (using OBI as example)

Currently 
- exact: /2009-11-02/obi.owl
  replacement: http://obi.svn.sourceforge.net/svnroot/obi/releases/2009-11-02/merged/merged-obi.owl

- exact: /2009-11-06/obi.owl
  replacement: http://obi.svn.sourceforge.net/svnroot/obi/releases/2009-11-06/merged/merged-obi.owl

Suggested:
- alias: repo=http://obi.svn.sourceforge.net/svnroot/obi/

- releasedate: 2009-11-02
  - releasebase: $(repo)/releases
  - release: obi.owl

- releasedate: 2009-11-06
[...]

PR #179, opened 2016-04-25
Update oostt.yml

mbrochhausen:
corrected previous mistake


jamesaoverton:
Thanks for the update. I'm working on reducing the distinction between "products" and "entries", to make it less confusing #106.

PR #183, opened 2016-05-06
Fix custom term browser for FLOPO

jamesaoverton:
- remove regex rule from flopo.yml
- add regex rule to obo.yml


jamesaoverton:
@leechuck: I forgot that custom term browsers have to be put in the higher-level `obo.yml`, not the project config file. Your flopo.owl and FLOPO_* PURLs should be working a few minutes after this pull request is merged.

leechuck:
Great, thank you! All seems to be working well.

PR #191, opened 2016-05-13
Update omrse.yml

hoganwr:
Fixed one PURL from old google code to GitHub and deleted another that didn't make sense


jamesaoverton:
Sorry: this needs to be updated now that #190 was merged. There should be an "update" button.

PR #194, opened 2016-05-17
Fixed VTO redirects.

balhoff:


jamesaoverton:
The changes should be live in about 20 minutes.

PR #195, opened 2016-05-17
Moved HAO to GitHub.

balhoff:


jamesaoverton:
Sorry, the `products` list should only be used for the top-level products: hao.owl and hao.obo

Please move the other two (`depictions.owl` and `arthropod-mappings.obo`) under `entries` using `exact` rules. Then they will appear as <http://purl.obolibrary.org/obo/hao/depiction.owl>, like this:

    - exact: /depictions.owl
      replacement: https://raw.githubusercontent.com/hymao/hao/master/depictions.owl
    - exact: /arthropod-mappings.obo
      replacement: https://raw.githubusercontent.com/hymao/hao/master/arthropod-mappings.obo

balhoff:
Thanks! I think I've fixed it how you suggest.

PR #198, opened 2016-05-20
Update omiabis.yml

mbrochhausen:
re-directed PURL

Question: Is the statement in line 68/69 redundant and should I delete that?

### DIFF
--- a/config/omiabis.yml
+++ b/config/omiabis.yml
@@ -4,7 +4,7 @@ idspace: OMIABIS
 base_url: /obo/omiabis
 
 products:
-- omiabis.owl: https://raw.githubusercontent.com/OMIABIS/omiabis-dev/master/BFO2%20omiabis/omiabis.owl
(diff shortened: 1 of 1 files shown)


jamesaoverton:
The `exact` entries are prefixed by the `base_url`, which is `/obo/omiabis` in this case. So `exact: /omiabis.owl` matches `http://purl.obolibrary.org/obo/ombiabis/omiabis.owl`. Is that what you want?

mbrochhausen:
Yes. I am wondering whether that is redundant and I could just delete the
exact: /omiabis.owl.

jamesaoverton:
You probably only want the `products` entry, which will match `http://purl.obolibrary.org/obo/omiabis.owl`.

mbrochhausen:
thanks!

On Fri, May 20, 2016 at 10:29 AM, James A. Overton <notifications@github.com

jamesaoverton:
I'm combining #198 and #199 into #200.

PR #199, opened 2016-05-20
Update omiabis.yml

mbrochhausen:
deleted redundant re-direct


jamesaoverton:
I'm combining #198 and #199 into #200.

PR #200, opened 2016-05-20
Update omiabis.yml

jamesaoverton:
Change product (#198), remove a redundant entry (#199).


jamesaoverton:
@mbrochhausen: The change should be live in about 20 minutes.

PR #207, opened 2016-06-16
edited d-acts PURLs

mbrochhausen:


jamesaoverton:
It would be better to add the 2016-04-22 PURL without removing the 2016-02-16 PURL. Is that possible?

mbrochhausen:
Yes, you are right. I'll do that.

PR #212, opened 2016-07-05
Update ido.yml

zhengj2007:
Add new release


cmungall:
I really recommend you don't check releases into master, but instead tag these. The github release mechanism is great for this. See https://github.com/cmungall/ontology-starter-kit

PR #213, opened 2016-07-06
Update dideo.yml

mbrochhausen:


cmungall:
I recommend not checking releases into master. Why not use the GitHub release mechanism

Spaces in folder names is also generally a bad idea

mbrochhausen:
Chris,

thanks, I will change the folder name.
Regarding the release mechanism is there an example of an OBO ontology doing that, so we can look into it?

Thanks,
Mathias

cmungall:
Have a look at the ontology starter kit, described here: 
https://github.com/monarch-initiative/monarch-disease-ontology

Also

https://github.com/pato-ontology/pato/
https://github.com/oborel/obo-relations/
https://github.com/obophenotype/porifera-ontology
https://github.com/obophenotype/ctenophore-ontology

mbrochhausen:
Thanks!

PR #218, opened 2016-07-08
Updated DIDEO.yml

mbrochhausen:
Updating DIDEO to use development folder instead of a release folder.  Also switching to using release prefix for future versions.


mbrochhausen:
Anything I need to do at this point?

PR #220, opened 2016-07-15
Update mpath.yml

leechuck:


jamesaoverton:
Hi @leechuck, cc @cmungall. A lot of workflows depend on OWL format, and do not support OBO format. I think we need an OWL version of MPATH, both here and in https://github.com/OBOFoundry/OBOFoundry.github.io/pull/289

cmungall:
Actually, users will continue to get the OWL via the fall-through mechanism, which goes to the central build at berkeley, which is built from the obo file (since this is what is registered in the [build tag](http://obofoundry.org/faq/what-is-the-build-field.html) on the [main mpath.md file](https://github.com/OBOFoundry/OBOFoundry.github.io/edit/master/ontology/mpath.md) ).

However, this is liable to confuse people! I think it's better to be explicit, and to directly choose one of two paths:

 1. mpath takes care of its own releases directly, uses robot or similar tools to deposit OWL files directly in the mapth GitHub repo with each release (of course the OWL can be placed anywhere, but GitHub is most convenient)
 2. mpath continues to use the obo build but explicitly registers the .owl file and points to it, and points the .obo to the same place, to avoid async issues between the two

1 involves a marginally higher initial overhead, but has lots of additional advantages, such as being able to use the GitHub release mechanism, being able to integrate ontology checks, etc. You can actually get started quite easily with https://github.com/cmungall/ontology-starter-kit, and this framework allows you to easily add in your own .groovy checks and so on.

cmungall:
Oh, I see now that this is coordinated with https://github.com/OBOFoundry/OBOFoundry.github.io/pull/289/commits/a29d29af56709ffca70062a46c6c8c83ff616a99, hence your comments @jamesaoverton which removes the build tag, which means the fall-through will end up on stale OWL.

OK, this is unusual. @leechuck could you inform us more as to your intent? I realize many mpath users don't use the OWL, but we do need the purl to resolve to the latest ontology for various things to work.

cmungall:
Maybe this is something to do with bioportal?

leechuck:
Yes, this was a request by @PaulNSchofield to help fix a problem with BioPortal; maybe @PaulNSchofield  can explain better what the problem is.

mcourtot:
@leechuk: what is the status on this and associated PR at https://github.com/OBOFoundry/OBOFoundry.github.io/pull/289 please?

leechuck:
@mcourtot I have not heard any news about problems with MPATH from Paul again, so I think it can be closed. Sorry for keeping it open so long.

mcourtot:
np @leechuck - thanks for taking care of this. Should I also close the associated PR?

leechuck:
Yes please, that would be great. Thank you for taking care of this!

PR #222, opened 2016-07-25
genepio initial configuration

Public-Health-Bioinformatics:
One question: I can't quite tell if github's genpio/imports/ folder should be referred to below with /src/ or without?  (Same question for the foodon config too).  Thanks.

https://raw.githubusercontent.com/GenEpiO/genepio/master/src/imports/

### DIFF
--- a/config/genepio.yml
+++ b/config/genepio.yml
@@ -1,0 +1,15 @@
+# PURL configuration for http://purl.obolibrary.org/obo/genepio
+
+idspace: GENEPIO
+base_url: /obo/genepio
+
+products:
+- genepio.owl: https://raw.githubusercontent.com/GenEpiO/genepio/master/genepio.owl
+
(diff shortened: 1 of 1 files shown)


cmungall:
It's up to you how you organize the files in your github repo.

My own preference is to have at the top level:

```
foo.{obo,owl}
imports/
    bar_import.owl
    ...
subsets/
    foo-basic.obo
```

And have everything under `src` be source only

This is how the starter kits sets things up

Public-Health-Bioinformatics:
Would be great to get this purl working soonish. :)

jamesaoverton:
We're usually faster, but I was offline for a few days. Please wait 30 minutes then double check that the PURLs are working as you expect.

Public-Health-Bioinformatics:
Looks like its all working, thanks!

PR #226, opened 2016-08-05
Update fbbt.yml

dosumis:
Syntax error in previous pull request.  This corrects it.  Change needed urgently.

### DIFF
--- a/config/fbbt.yml
+++ b/config/fbbt.yml
@@ -31,7 +31,10 @@ entries:
   replacement: http://www.ontobee.org/browser/rdf.php?o=FBbt&iri=http://purl.obolibrary.org/obo/
 
 - prefix: /vfb/
-  replacement: https://raw.githubusercontent.com/VirtualFlyBrain/VFB_owl/master/src/owl/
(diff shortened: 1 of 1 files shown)


dosumis:
Not clear to me why this is failing:

FAIL	/obo/fbbt/vfb/vfb.owl	302	http://virtualflybrain.org/owl/vfb.owl	302	

With the old Purl server system, it was possible to have an exact redirect under a partial.  Is this still the case?

cmungall:
I'm at a meeting and James is on a plane.

It looks like James corrected the problem by having the `exact:` precede the `prefix:`. But this had a typo, David corrected this, but reintroduced the ordering issue (and a conflict).

I can try and look at this later I don't want to rush this and introduce further confusing. @dosumis perhaps just close this, make a new PR, with the desired URLs, but make the exact precede the prefix (I believe it's in order of priority for matching)

jamesaoverton:
@dosumis: Is it working as expected now?

dosumis:
> @dosumis: Is it working as expected now?

Yep.  That's perfect.  Thanks!

PR #229, opened 2016-08-14
Update to handle new versioning 

Robbie1977:
VFB is moving to github tag/version revisions rather than the folder system previously used.

I've added redirects for historical versions as well as new versions.

Larger ontologies are now being gzipped hence the redirects for specific OWL files.

Requested by @dosumis in https://github.com/VirtualFlyBrain/VFB_owl/issues/29

### DIFF
--- a/config/fbbt.yml
+++ b/config/fbbt.yml
@@ -33,31 +33,31 @@ entries:
 - exact: /vfb/vfb.owl
   replacement: http://virtualflybrain.org/owl/vfb.owl
   
-- regex: ^/vfb/(2016-0[0-7]-[0-9][0-9])/vfb\.owl&
+- regex: ^/obo/fbbt/vfb/(2016-0[0-7]-[0-9][0-9])/vfb\.owl&
   replacement: https://raw.githubusercontent.com/VirtualFlyBrain/VFB_owl/Archive/src/owl/$1/vfb.owl.gz
   tests:
   - from: /vfb/2016-01-21/vfb.owl
     to: https://raw.githubusercontent.com/VirtualFlyBrain/VFB_owl/Archive/src/owl/2016-01-21/vfb.owl.gz
   
-- regex: ^/vfb/(2016-0[0-7]-[0-9][0-9])/fbbt_(.*)\.owl$
+- regex: ^/obo/fbbt/vfb/(2016-0[0-7]-[0-9][0-9])/fbbt_(.*)\.owl$
   replacement: https://raw.githubusercontent.com/VirtualFlyBrain/VFB_owl/Archive/src/owl/$1/fbbt_$2.owl
   tests:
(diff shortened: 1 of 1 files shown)


cmungall:
I'm not sure if the tests are failing because the regexs aren't supported, or if it's just that the test suite is making assumptions.

Btw, I always mint versionIRIs that have the string `/releases/` in them, e.g. `/obo/uberon/releases/YYYY-MM-DD/`. I started that back when we had the old OCLC system where you couldn't do regexes at all. Overall I find it makes configuration easier, but YMMV

Robbie1977:
happy to swap the ordering around but trying to match what has already been used, however the issue is with the added subdirs which could be resolved by moving the OWL files to the base folder as a last result. I'll have another go at getting the syntax to work (tests are a great help) just need to figure out your syntax differences.

Any pointers on what is failing in the syntax?

Robbie1977:
All fixed!

@dosumis if you want to take @cmungall suggestion above (switch to /release/ subdir) speak now or forever hold your peace?

Robbie1977:
@dosumis I've just added it anyway so it's there if we chose to use it in future, I've not bothered making if backwardly compatible.

Robbie1977:
I'm all finished please merge when you have time.

jamesaoverton:
Double checking: Are you sure that you want various `*.owl` PURLs to redirect to `.owl.gz` files? Protege seems to handle it in a quick test (that was a pleasant surprise), but other tools may break.

dosumis:
> Are you sure that you want various *.owl PURLs to redirect to .owl.gz

This is intentional.  The current VFB owl file and any dataset smaller than 50M (all so far) are all available as plain owl files.  Archival copies of the full VFB owl file are only available as .gz.  The latest versions of Protege and the OWL-API are fine with this.

dosumis:
> @dosumis if you want to take @cmungall suggestion above (switch to /release/ subdir) speak now or forever hold your peace?

That was also my suggestion on the VFB_owl ticket (!)  If I understand the new tag system correctly, this means that redirects for any of our OWL files will go to the tag version with the appropriate date:

- regex: ^/obo/fbbt/vfb/(20.*)/$
  replacement: https://raw.githubusercontent.com/VirtualFlyBrain/VFB_owl/$1/src/owl/
  tests:
  - from: /vfb/2016-07-12/
    to: https://raw.githubusercontent.com/VirtualFlyBrain/VFB_owl/2016-07-12/src/owl/

But see http://purl.obolibrary.org/obo/fbbt/vfb/2016-07-12

(perhaps I'm being impatient and not yet up.  Perhaps one of other regex redirects is over-riding in this case?

Robbie1977:
@dosumis that is due to the final / missing which wouldn't be expected for referencing of the owl files. This can be allowed for if required?

PR #246, opened 2016-09-28
Update iao d-acts

utecht:
Moving d-acts to different release/development layout and future releases via tagging.


jamesaoverton:
I'm not sure it's a good idea to redirect `/release/` for all of IAO to a d-acts specific directory. Should it be `prefix: /d-acts/release/` instead?

utecht:
Thank you for catching that I'll update it and issue a new pull request.

PR #252, opened 2016-10-18
Updated links to current versions

cooperl09:


jamesaoverton:
The ontology links are good.

Do you have a new tracker that we can use to replace the old tracker PURL? PURLs should be permanent, so I'd prefer to replace the `/tracker` entry rather than remove it.

cooperl09:
Hi James,
I added the link to the tracker back- with the updated one at GH
L

config/eo.yml
```diff
@@ -4,12 +4,13 @@ idspace: EO
...
+- eo.OBO: https://raw.githubusercontent.com/Planteome/plant-environment-ontology/master/plant-environment-ontology.obo
 
 term_browser: ontobee
 example_terms:
 - EO_0007359
 
 entries:
 - exact: /tracker
-  replacement: https://sourceforge.net/tracker/?atid=5010796&group_id=76834&func=browse
+- replacement: https://github.com/Planteome/plant-environment-ontology/issues
```
jamesaoverton: This line should start with two spaces, not a hyphen and a space.

Sorry to nitpick, but computers are *very* literal-minded.

cooperl09: Ok, will change, but I just copied and pasted what was in the file previously, then replaced the link to the tracker....

PR #258, opened 2016-10-27
Added GO_REFs regex

mcourtot:
Added regex for GO_REFs as per https://github.com/geneontology/go-site/issues/214

### DIFF
--- a/config/go.yml
+++ b/config/go.yml
@@ -38,16 +38,21 @@ entries:
 - prefix: /about/
   replacement: http://www.ontobee.org/browser/rdf.php?o=GO&iri=http://purl.obolibrary.org/obo/
 
-- prefix: /
-  replacement: http://geneontology.org/ontology/
-  
-# Terms for GO REFs
-# https://github.com/geneontology/go-site/blob/master/metadata/gorefs/goref-0000001.md
-# should point to
+  # Terms for GO REFs
 # http://purl.obolibrary.org/obo/go/references/0000001
+# should point to
+# https://github.com/geneontology/go-site/blob/master/metadata/gorefs/goref-0000001.md
+
+
 - regex: ^/obo/go/references/(\d+)$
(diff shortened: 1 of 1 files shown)


mcourtot:
hm - there is an extra /go/ in the test which causes the fail. let me try and see if I can commit over my PR...

config/go.yml
```diff
@@ -40,4 +40,14 @@ entries:
 
 - prefix: /
   replacement: http://geneontology.org/ontology/
+  
+# Terms for GO REFs
+# https://github.com/geneontology/go-site/blob/master/metadata/gorefs/goref-0000001.md
+# should point to
+# http://purl.obolibrary.org/obo/go/references/0000001
+- regex: ^/references/(\d+)$
```
jamesaoverton: Confusingly unlike the others entry types, `regex` has to match the full path. It should be: `^/obo/go/references/(\d+)$`

config/go.yml
```diff
@@ -40,4 +40,14 @@ entries:
...
+  replacement: https://github.com/geneontology/go-site/blob/master/metadata/gorefs/goref-$1.md
+  tests:
+  - from: /references/0000001
```
jamesaoverton: This test should be right now -- that's the confusing part. Sorry.

mcourtot:
Hm - still failing tests. Shall I just merge and hope for the best :) Maybe naive question, but how can the test succeed before the PURL is setup? Is it expected to fail at this stage?

jamesaoverton:
Now the problem is that that regex entry comes after the `prefix: /` entry. The rules are matched in order, and `prefix: /` will match everything starting with `http://purl.obolibrary.org/obo/go/`.

Move the regex somewhere before `prefix: /` and it should be good.

mcourtot:
Yes - all passed (finally!). Thanks again for the help. Can you merge if it all looks ok to you as well?

jamesaoverton:
Sorry for the hassle. I admire your perseverance! :^)

PR #260, opened 2016-11-16
Do not merge - Added PDRO file

mcourtot:


mcourtot:
Help with the failed checks please?
Also please do not merge yet, the URIs in the OWL file are not of the right format, see https://github.com/OBOFoundry/OBOFoundry.github.io/issues/283#issuecomment-260918073

jamesaoverton:
Not your fault. Your changes look fine, but the test system is failing long before it looks at the new file. I think this is a problem with Travis, and we don't actually want to merge yet, so I'm going to leave it alone for a while and try again later.

mcourtot:
Great, thanks for checking.
Do you know what would happen if we were to merge a file with 'improper' URIs? I assume it would try and fetch the whole file every time, or just be unable to split the fragment after the base URL?

jamesaoverton:
Most configuration errors would only break PURLs for that namespace, and usually the problem is that the file just can't be found. But a bad regex entry could break all of our PURLs. Tests are good.

Travis seems to be back to normal now.

mcourtot:
URIs seem to have been fixed in the https://raw.githubusercontent.com/OpenLHS/PDRO/master/PDRO.owl file. Merging PR.

Issue #267, opened 2016-11-29
Add schema checker for YAML files


kltm:
We recently had an outage due to improper redirects. There should be schema checking for all files.

https://github.com/geneontology/minerva/issues/77

https://github.com/OBOFoundry/purl.obolibrary.org/commit/9994118a79124ffa80f128f7921f0698f284a429

jamesaoverton:
Every YAML config file is checked against the [config.schema.yml](https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/tools/config.schema.yml#L70), and in a few places it checks the first part of a URI. Unfortunately the checks don't catch this particular problem. Suggestions are welcome.

cmungall:
This was the offending PR:

https://github.com/OBOFoundry/purl.obolibrary.org/pull/264

It passed Travis-CI. Note a schema check wouldn't help here as it adhered to the schema.

It's also hard to test for redirect URLs in the general case, since a partial redirect is not guaranteed to work without a suffix.

We could do some preliminary checking on the URL to avoid errors like `http:///` (three slashes) but it's hard to anticipate the dumb errors that users like @cmungall might make.

We do have a very nice test mechanism. We would do more to encourage (force?) cfgs to have tests, especially for infrastructure important ontologies like this one. That would have prevented this.

kltm:
kwalify allows a regexp on fields; what are you using?

jamesaoverton:
We're using kwalify with a regex, but the particular regex we're currently using did not reject that triple-slash: https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/tools/config.schema.yml#L70

kltm:
We now have something like this for something similar:
```regexp
/^((ht|f)tp(s?)\:\/\/\w[\/\.\-\:\w]+)|(GOC\:[\w\_]+)$/
```
Maybe you could have something like:
```regexp
/^(https?|ftp)\:\/\/\w[\/\.\-\:\w]+$/
```

kltm:
For at least the slashes, I have put in a PR here: https://github.com/OBOFoundry/purl.obolibrary.org/pull/268

I also noticed a few places that could probably be improved, but I'm unsure if there are corner cases that would not work. For example the schema check for "products" has _anything_ allowable, but it seems like it could at least be matched by something like:
```regexp
[...]

PR #268, opened 2016-11-29
more correct url checker; corrected comment examples; work on #267

kltm:


jamesaoverton:
Thanks @kltm!

@cmungall: Travis is saying that all the current PURLs pass this new regex. I'm confident in the test coverage and happy to merge. Do you agree?

PR #270, opened 2016-12-01
temp location for vt,

cmungall:
See https://github.com/OBOFoundry/OBOFoundry.github.io/issues/257


cmungall:
cc @caripark - I placed a vt.obo and vt.owl in a temporary github location for now until we figure out getting you set up.

PR #313, opened 2017-03-10
adding core, fixing imports path

Public-Health-Bioinformatics:
Currently the /imports/ folder path is wrong, causing it to fail to load in OLS and others.  This fixes.  As well, ontology now has a foodon_core.owl file.


jamesaoverton:
PURLs are supposed to be permanent. You shouldn't just remove the `foodon.obo` PURL, it should point to something, even if it's an older version.

Public-Health-Bioinformatics:
Ah, of course. I'll rectify that.

jamesaoverton:
Travis isn't running the checks, but I'll merge anyway.

PR #329, opened 2017-06-04
Update po config to include planteome.owl

cmungall:
planteome.owl is an importer ontology for the Planteomeproject, covering all plants.

The proposal is to use this PURL
http://purl.obolibrary.org/obo/po/planteome.owl

Underneath the 'po' plant anatomy reference ontology (even though planteome is more than anatomy, this seems a reasonable place for this)

See: https://github.com/OBOFoundry/OBOFoundry.github.io/issues/405

Fixes https://github.com/Planteome/planteome-core/issues/25


cmungall:
cc @cooperl09 @elserj

PR #336, opened 2017-06-30
Revert to previous image as latest seems to be breaking ansible

cmungall:
See: https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch


cmungall:
Attempt to fix #335

PR #342, opened 2017-07-19
Remove global redirects for HAO purls

cmungall:
These should be handled in the main HAO file


cmungall:
@balhoff - FYI HAOs still resolve to your desired location, so this block I deleted was redundant.

PR #355, opened 2017-09-15
Adding redirect for images

srobb1:


jamesaoverton:
It looks like there's an extra space at the beginning of line 25.

PR #362, opened 2017-10-19
Get alternative products and versions to resolve

nataled:


config/pr.yml
```diff
@@ -9,14 +9,28 @@ base_redirect: http://proconsortium.org
...
+- pr-asserted.owl: ftp://ftp.proconsortium.org/databases/ontology/pro_obo/pro_nonreasoned.owl
+- pr-asserted.obo: ftp://ftp.proconsortium.org/databases/ontology/pro_obo/pro_nonreasoned.obo
 
 term_browser: custom
 
 entries:
+- prefix: /release/
+  replacement: ftp://ftp.proconsortium.org/databases/ontology/pro_obo/old_files/release_
+  tests:
+  - from: /release/52.0/pr.obo
```
jamesaoverton: Unfortunately this won't work. You'll get `ftp://ftp.proconsortium.org/databases/ontology/pro_obo/old_files/release_52.0/pr.obo` without `_reasoned`.

nataled: I was afraid of that. Would it be possible to declare--prior to the above--something like
- exact: pr.owl
   replacement: pro_reasoned.owl
(here hoping that the lack of slash will allow just the last part to get replaced)

nataled:
Awesome. I tried to avoid the regex method since there was some warning about it in the docs. The only additional thing needed is to deal with the unfortunate legacy versioning. Apparently our versions had a trailing period attached, so
52.0.
   instead of
52.0

I believe I can add to your regex a \.* after the \d+ and before the parenthesis to take care of this? And for the two old cases where we have three period-separated digit fields I can just hard code them. We haven't done that in years.

jamesaoverton:
Ok, I added an optional trailing period. Does it look right now?

nataled:
So far as I can tell. Thanks for the assist!

PR #377, opened 2017-11-06
Update pco.yml

ramonawalls:
Fix redirect URL for merged file


config/pco.yml
```diff
@@ -15,7 +15,7 @@ entries:
 
 - exact: /pco_merged_inferred.owl
   replacement: https://raw.githubusercontent.com/PopulationAndCommunityOntology/pco/master/src/ontology/pco_merged_inferred.owl
-
+
```
jamesaoverton: All this commit did was add some whitespace. I don't think that's what you intended.

ramonawalls: No not really. :) Thanks @jamesaoverton

PR #380, opened 2017-11-14
Try full path for kwalify

jamesaoverton:


jamesaoverton:
This failed: bash still can't find kwalify.

Issue #290, opened 2017-01-12
Travis false positive on schema validation


jamesaoverton:
Reported by @alanruttenberg in #287. The `geno.yml` file did not include the required `term_browser` key, which caused `make validate` to fail in Alan's local environment. Validation also failed on the PURL server, blocking deployment (which is good), but nobody noticed until Alan manually checked his IAO changes #286.

However all tests passed on Travis (<https://travis-ci.org/OBOFoundry/purl.obolibrary.org/builds/190343322>). I'm very unhappy about this, because I was trusting Travis to prevent bad merges.

jamesaoverton:
I just found another one of these: a duplicate entry passed validation on Travis but not on the server. Fixed by #379.

Now I see that `kwalify` is failing on Travis without causing the build to fail: `bash: kwalify: command not found` https://travis-ci.org/OBOFoundry/purl.obolibrary.org/builds/302132357#L577 It looks like this was the case for the original bug report.

Working on this in #380.

jamesaoverton:
Well, now we know three things that don't work...

jamesaoverton:
Ran into this again today. :(

kltm:
@jamesaoverton If I'm reading this correctly, I think this can be easily worked around. You may want to look at the travis file we use for reusable data:
https://github.com/reusabledata/reusabledata/blob/master/.travis.yml

jamesaoverton:
Thanks @kltm. I would very much like to see this fixed, but I’m going on vacation for a few weeks. If you have time to try the fix, that would be great.

kltm:
@jamesaoverton At this point, I'm fairly unfamiliar with how the site/testing hangs together, but I should be able to bug @cmungall enough to try a PR.

cmungall:
@kltm - it's all self-contained ansible so you should be able to get it running without knowing the ins and outs

Let me see if I understand the issue first, I'll try and restate

The validate step requires kwalify. If kwalify is not found, then validation auto-passes. This is not good as it leads to different behavior between travis and production.

An example from the above build where kwalify silently fails on Travis:

```
72.72s$ sudo make all test
rm -rf temp tests
kwalify -f tools/config.schema.yml config/*.yml \
	| awk '{print} /INVALID/ {status=1} END {exit status}'
bash: kwalify: command not found
mkdir backup
mkdir -p temp/obo/aeo
```

This is unusual as we expect the build to fail if kwalify is not found. We should fix the makefile
[...]

PR #381, opened 2017-11-14
Try --no-user-install for kwalify

jamesaoverton:
See #290


jamesaoverton:
This experiment failed.

PR #391, opened 2018-01-11
Fix imports path

cmungall:


cmungall:
https://github.com/monarch-initiative/mondo-build/issues/56

Issue #394, opened 2018-02-02
Update NBO Pointer to New Build location


reality:
Pull request incoming

cmungall:
Looks like it's done?

PR #395, opened 2018-02-02
Update nbo.yml to point to new ontology location [#394]

reality:


jamesaoverton:
The change should be live in < 20 minutes. If not, let us know.

reality:
Seems to have worked fine. Thanks for the quick update!

PR #401, opened 2018-02-06
Update doid.yml

lschriml:
Adding additional file to DO's perl.


jamesaoverton:
All `products` should start with the project's IDSPACE, in this case `doid`. They end up as `http://purl.obolibrary.org/obo/IDSPACE.XXX`. We have strict rules about this top-level.

Under `http://purl.obolibrary.org/obo/IDSPACE/` you are not limited in this way. Use `entry` for these.

In this case I suggest `http://purl.obolibrary/obo/doid/HumanDO.obo` would would be a new `entry`. If that's ok, I can make the update.

lschriml:
James, good solution. 
Thank you,
Lynn

PR #402, opened 2018-02-06
Update doid.yml

lschriml:
update sent of DO Products


jamesaoverton:
Let's take care of this in #401

PR #412, opened 2018-03-09
Update peco.yml

cmungall:
add path for imports and subsets


cmungall:
@marieALaporte check your email for an invite to the OBO community group

PR #416, opened 2018-03-14
Update pr.yml

nataled:
Not all products currently produced were available in past releases. Therefore the versioned PURLs need to be customized to handle the differences. Prior to release 43.0 only non-reasoned OBO files were produced but after that there are both OBO and OWL reasoned and non-reasoned files. 

Also, there has been a large-scale move and rename of many PRO files so nearly all PURLs had to be revised.

### DIFF
--- a/config/pr.yml
+++ b/config/pr.yml
@@ -24,7 +24,7 @@ entries:
   - from: /9.0/pr.obo
     to: https://research.bioinformatics.udel.edu/PRO/data/release_9.0/pro_nonreasoned.obo
 - regex: (?i)^/obo/pr/([123]\d\.\d+)\.?/pr.obo$
-  replacement: https://research.bioinformatics.udel.edu/PRO/data/release_$1/pro_non_reasoned.obo
+  replacement: https://research.bioinformatics.udel.edu/PRO/data/release_$1/pro_nonreasoned.obo
   tests:
   - from: /22.0/pr.obo
     to: https://research.bioinformatics.udel.edu/PRO/data/release_22.0/pro_nonreasoned.obo
@@ -93,9 +93,9 @@ entries:
   replacement: https://research.bioinformatics.udel.edu/PRO/data/release_$1/pro_nonreasoned.obo
   tests:
   - from: /51.0/pr-asserted.obo
-    to: ftp://ftp.proconsortium.org/databases/ontology/pro_obo/old_files/release_51.0/pro_nonreasoned.obo
+    to: https://research.bioinformatics.udel.edu/PRO/data/release_51.0/pro_nonreasoned.obo
   - from: /51.0./pr-asserted.obo
-    to: ftp://ftp.proconsortium.org/databases/ontology/pro_obo/old_files/release_51.0/pro_nonreasoned.obo
+    to: https://research.bioinformatics.udel.edu/PRO/data/release_51.0/pro_nonreasoned.obo
 
 - regex: (?i)^/obo/pr/(4[3456789]\.\d+)\.?/pr-asserted.owl$
   replacement: ftp://ftp.proconsortium.org/databases/ontology/pro_obo/old_files/release_$1/pro_nonreasoned.owl
--- a/config/pr.yml
+++ b/config/pr.yml
@@ -7,56 +7,111 @@ base_url: /obo/pr
 base_redirect: http://proconsortium.org
 
 products:
-- pr.owl: ftp://ftp.proconsortium.org/databases/ontology/pro_obo/pro_reasoned.owl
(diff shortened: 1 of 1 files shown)


jamesaoverton:
As far as I can tell, these regexes are doing what the comments say they should. And the tests pass.

I'm worried that it will be hard to maintain this. Do you mind if I try to simplify it?

cmungall:
We don't have an explicit rule for this but we should: The only top levels allows should be `$ONT.$FMT`, `$ONT` and anything under `$ONT/`. So for pr-asserted, make this `pr/pr-asserted.$FMT`.

Also we (OBO) need to standardize ASAP the naming conventions for different products. @balhoff  proposes `raw` for (1) no-inference (2) no-imports (3) no-external-classes

balhoff:
Lately I was thinking `atomic` rather than `raw`. Need to make some time to create some tools for creating & working with those.

nataled:
You can try if you want, but there is no maintenance required (assuming 
by 'maintain' you mean new PURLs to handle the same type of regex), at 
least not until we reach 3-digit major release numbers. I actually 
already have the regex needed for that, but since I cannot test it I 
left it out. I also have date-type PURLs in the pipeline, but we aren't 
ready for those just yet.

jamesaoverton:
"Maintenance" in the software development sense: Someday there will be a bug, and somebody will have to read this and understand it well enough to fix the bug or be sure that the bug is elsewhere.

The key distinction is between versions 0-42 and 43+, but the 0-42 case has three special cases (0-9,10-39,40-42). I want to try and collapse those three special cases into a single case using regex alternation. We'll see if it's harder or easier to read that way.

nataled:
That's a holdover from an earlier misconception on my part. The PURL you 
indicate is already in place. I just haven't deleted that other one yet 
(I wanted to make sure everything worked first).

nataled:
Oh, gotcha. Sure, that's fine. I think it should also be possible, in 
the 43+ case, to collapse the extensions too, correct? I mean the format 
specifier (obo vs owl).

nataled:
Hi James,

It looks like none of the changes I made went into effect. I can think of two reasons. One is that it is being held up while you work on the simplification (I believe that this is not actually the case). The other is that there is some problem with the pull request. When I first made the changes, there was one typo and one not-fixed PURL destination. Accordingly, the build failed. I checked the Travis log, found the errors, and went to fix them, but there doesn't seem to be a way to get back to the editing page for the pull request. I therefore made another pull request, which passed all checks, but all it seemed to do was merge into the previous failed branch. In any case, my attempts to verify the changes took effect indicated that they did not, and when I went to a 'clean' page to look at the PURL YAML it's clear that none of the changes took hold. Any suggestions? I can't even get back to my first branch 'nataled-patch-2'

jamesaoverton:
The delay is my fault. I have a simplified version that I'll post in a minute for you to check.

But first I have a question. The only FTP in the new version is here:

```
- regex: (?i)^/obo/pr/(4[3456789]\.\d+)\.?/pr-asserted.owl$
  replacement: ftp://ftp.proconsortium.org/databases/ontology/pro_obo/old_files/release_$1/pro_nonreasoned.owl
  tests:
  - from: /45.0/pr-asserted.owl
    to: ftp://ftp.proconsortium.org/databases/ontology/pro_obo/old_files/release_45.0/pro_nonreasoned.owl
  - from: /45.0./pr-asserted.owl
    to: ftp://ftp.proconsortium.org/databases/ontology/pro_obo/old_files/release_45.0/pro_nonreasoned.owl
```

Is this correct or an oversight?

nataled:
Oversight, though I thought I changed all links to ftp when I fixed the typo. Apparently browser page searches don't work well within that .md editor tool. When I searched for 'ftp' using my browser (Firefox) it returned 'not found' even though there were occurrences (apparently).

jamesaoverton:
No problem. I made a new PR #420. If you agree that it's correct and easier to understand, we'll merge it and close this PR.

Here's the new `pr.yml` by itself: https://github.com/OBOFoundry/purl.obolibrary.org/blob/7107a6ce3cd8f6a7fc87d335b3db3a9fc55adea3/config/pr.yml

nataled:
I am no longer sure where we are with this. I see a note indicating that this branch has conflicts. I also see that whatever we had under "entries:" is now gone. I'm guessing that whatever change was made has not gone into effect, because PURLs for entries still work. I still have more PURLs to add but the uncertainty about changes already made is keeping me from doing so.

jamesaoverton:
Sorry. #420 supersedes this and was merged successfully. I'll close this PR. Please start with `master` for the new changes you mentioned.

PR #420, opened 2018-03-15
PR updates, revised version of #416

jamesaoverton:


nataled:
Looks good, even though I don't personally find it human readable (not an issue as that's simply a function of time spent staring at it). One question though:

Chris mentioned that we shouldn't have a http://purl.obolibrary.org/pr-asserted.obo and that it should have an extra bit of /pr added. So far as I can tell, this is solved by the regex purls. However, I see these:

- exact: /pr-asserted.obo
  replacement: https://research.bioinformatics.udel.edu/PRO/data/current/pro_nonreasoned.obo
- exact: /pr-asserted.owl
  replacement: https://research.bioinformatics.udel.edu/PRO/data/current/pro_nonreasoned.owl

Are these the cause of the 'problem'? Should they be revised to '/pr/pr-asserted.obo' etc? Or maybe the problem lies under the 'products' part?

jamesaoverton:
Chris was referring to this part:

```
products:
- pr.owl: https://research.bioinformatics.udel.edu/PRO/data/current/pro_reasoned.owl
- pr.obo: https://research.bioinformatics.udel.edu/PRO/data/current/pro_reasoned.obo
- pr-asserted.owl: https://research.bioinformatics.udel.edu/PRO/data/current/pro_nonreasoned.owl
- pr-asserted.obo: https://research.bioinformatics.udel.edu/PRO/data/current/pro_nonreasoned.obo
```

`products` are the primary file PURLs, and we require every ontology to have an OWL product corresponding to their ID. We should also prohibit projects from using names outside their ID for products, because that amounts to using up another ID. `pr-asserted` is a case in point.

The `pr-asserted` products were added on 2016-10-03 with this commit: https://github.com/OBOFoundry/purl.obolibrary.org/commit/d04677520313240a253a9ca948cf08bc342a5e9c

There should have been a test to prevent that.

We won't change them now.

nataled:
As the offending PURLs have never been advertised, they *could* be deleted if desired. At the time I added those 'asserted' guys I thought the list simply represented files and not PURLs per se. I leave it up to you. I guess though that we would need to add

- pr/pr-asserted.owl
- pr/pr-asserted.obo

If I understand the issue correctly.

jamesaoverton:
I dunno. What do you think @cmungall ?

cmungall:
Yes, you should use pr/pr-asserted.owl

If you are concerned about people having hardcoded the old URL, then we can leave it up for a while with appropriate 30x redirect.

nataled:
I have no such concern. As mentioned, the ONLY place that URL appears is in the PURL file. The page that describes the PRO PURLs has it correctly indicated.

PR #422, opened 2018-03-16
Remove extra PR products

jamesaoverton:
See discussion on #420

### DIFF
--- a/config/pr.yml
+++ b/config/pr.yml
@@ -9,8 +9,6 @@ base_redirect: http://proconsortium.org
 products:
 - pr.owl: https://research.bioinformatics.udel.edu/PRO/data/current/pro_reasoned.owl
 - pr.obo: https://research.bioinformatics.udel.edu/PRO/data/current/pro_reasoned.obo
-- pr-asserted.owl: https://research.bioinformatics.udel.edu/PRO/data/current/pro_nonreasoned.owl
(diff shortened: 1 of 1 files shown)


jamesaoverton:
These extra products are not needed by PR. They were only added because of confusion over the meaning of `products`. That's entirely my fault.

We need an automated check for this rule. This is the only exception that I'm aware of. Life will be easier if we remove the exception, and sooner is better.

Ok @nataled ?

nataled:
That is fine. Bear in mind that we DO have such files, but the PURL is incorrect. I presume that would be handled somewhere below, by 'exact' or something.

nataled:
Any idea why PRO PURLs for individual terms still work? The part of the PURL yaml for these is missing. I know that the old purl server (before there was purl.obolibrary.org) had been set up and probably still works. Maybe there is failover to that server instead?

jamesaoverton:
Im not sure what you mean. PRs just show the changes. The current version of `pr.yml` on the `master` branch has everything: https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/pr.yml

cmungall:
@nataled it's generic, no need to register terms

nataled:
So if I needed to specify a different destination, how would I do that? It used to be in this file, as far as I remember.

nataled:
I mean there was a regex to do it.

PR #423, opened 2018-03-17
Added handling of date-type PURLs

nataled:


jamesaoverton:
The regex looks correct, although `\d\d\d\d-\d\d-\d\d` might be sufficient.

One problem is that this `replacement: https://hershey.dbi.udel.edu/natale-cgi/getPRO?request=$1` is completely different from the `to` URLs in the tests that follow it.

Another problem is that a regex alone can't transform `pr|pr-asserted` to `pro_reasoned` and `pro_nonreasoned` respectively. That requires two rules, as elsewhere in `pr.yml`.

nataled:
My understanding is that the tests should point to the ultimate destination. Here's what happens:
1) The PURL is replaced by the getPRO link
2) The getPRO script converts the date-type PURL to a major-minor type PURL
3) The major-minor PURL points to the file indicated in the tests.

Should I stop at (1) ?

Regarding the pr|pr-asserted part, that's only to indicate what's allowed to match the PURL. Technically the PURL isn't doing any conversion of the pr|pr-asserted to reasoned vs nonreasoned. That's handled by getPRO (indirectly, since all getPRO does is return another PURL, which is then handled by the 'elsewhere in pr.yml').

One other thing that's been a real thorn in my side when editing these. When I get a failure, getting back to the file to edit isn't trivial, and once I edit it I cannot find how to do another pull request. Is it even possible at my end?

jamesaoverton:
The PURL system really doesn't do much (by design). It's just a web server that accepts a URL and returns a new URL as an HTTP 30X Redirect. Browsers and other tools will automatically follow the redirect. There may be many redirects before you get to the resource you want, but this system only handles one step in that chain.

So for a test you're not specifying the final URL, only the next step. The tests just check for the one step that the PURL system handles.

GitHub Pull Requests correspond to a Git branch. You can commit further changes to the branch, push them, and they'll be included in the Pull Request.

You can use the web interface to do that: From the Pull Request page, click the Files Changed tab. For each changed file there's a View button and a pencil icon for editing. When you're done editing, commit your changes to the branch (`nataled-patch-3` in this case) and they'll be added to the Pull Request.

nataled:
Okay, thanks. I was hoping I could then run the checks, but seems no luck there. The complicated date was to prevent crazy things like 1047-18-75 :)

jamesaoverton:
One of the tests says `017-02-01` but it should be `2017-02-01`. Then it should work.

PR #426, opened 2018-04-16
Update wbphenotype.yml

dosumis:
Adding redirect for importer file that pulls in wbphenotype and formalization.


cmungall:
Any reason to put this in the /wbphenotype space rather than upheno?

cmungall:
I believe WBpheno is now managing their axioms with the ontology Nico set up for them?

matentzn:
Please delete this pull request, we are changing everything to the new system now. :) Thanks

PR #428, opened 2018-05-08
Update .htaccess to add a /meta/ route

### LINKED ISSUE #427: PURL for registry metadata
we make these available

https://raw.githubusercontent.com/OBOFoundry/OBOFoundry.github.io/master/registry/ontologies.yml
https://raw.githubusercontent.com/OBOFoundry/OBOFoundry.github.io/master/registry/ontologies.ttl

I'm getting increasing requests for these but this doesn't seem like a good URL to give out, I'd like to give out a PURL

What do we think of `http://purl.obolibrary.org/meta/` as the base URL? With fragments like ontologies.yml, ttl. We could also have void dumps in future.

If we agree I can look into making a PR on the code in this repo

cmungall:
Not tested!

If works, fixes #427

### DIFF
--- a/www/.htaccess
+++ b/www/.htaccess
@@ -6,3 +6,8 @@ RedirectMatch temp ^/obo/(.*)$ http://purl.oclc.org/obo/$1
 # Redirect anything else to the GitHub page.
 RedirectMatch temp ^.*$ https://github.com/OBOFoundry/purl.obolibrary.org/
 
+### Meta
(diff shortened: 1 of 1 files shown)


cmungall:
ping @jamesaoverton

jamesaoverton:
This will make PURLs like `http://purl.obolibrary.org/meta/ontologies.ttl` (not `registry.ttl` as discussed on #427. I think that's fine.

PR #441, opened 2018-07-24
Create dpo.md

jamesaoverton:
See https://github.com/OBOFoundry/OBOFoundry.github.io/pull/665

This creates the main `dpo.owl` PURL. I see there are other releases in <https://github.com/FlyBase/flybase-controlled-vocabulary/tree/master/releases>. We can add PURLs for those too, if desired.

### DIFF
--- a/config/dpo.md
+++ b/config/dpo.md
@@ -11,6 +11,6 @@ products:
 
 term_browser: ontobee
 example_terms:
-- FBdv_00005370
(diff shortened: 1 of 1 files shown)


dosumis:
Thanks for this.

One small change needed. The example term comes from an import, rather than DPO.  Please use 'FBcv:0000670' instead.

PR #463, opened 2018-11-21
Update wbphenotype.yml

chris-grove:


cmungall:
@jamesaoverton I think we want to discourage top-level products like myown-artefact, instead the top level should just be `.FMT`, and additional artefacts should go in a subdir.

However, I don't think we explicitly wrote this down, and the rule may just be in my head?

config/wbphenotype.yml
```diff
@@ -4,10 +4,23 @@ idspace: WBPhenotype
...
 
 term_browser: ontobee
 example_terms:
 - WBPhenotype_0000886
 
+entries:
+- prefix: /releases/
+  replacement: https://raw.githubusercontent.com/obophenotype/c-elegans-phenotype-ontology/v
+- prefix: /tracker/
+  replacement: https://github.com/obophenotype/c-elegans-phenotype-ontology/issues
```
jamesaoverton: I'm worried that this needs a trailing slash to match the prefix.

config/wbphenotype.yml
```diff
@@ -4,10 +4,23 @@ idspace: WBPhenotype
...
+- prefix: /about/
+  replacement: https://github.com/obophenotype/c-elegans-phenotype-ontology
```
jamesaoverton: I think this also needs a trailing slash.

jamesaoverton:
@cmungall This raises the question: What are we counting as a `product` now? We've always reserved it for files with the same name as the IDSPACE. I can see making an exception for `base`. If anything goes, then we might conflict with other IDSPACES (although I'm not really worried about this particular config file).

cmungall:
I may have introduced some terminological inconsistency, as over on the obo metadata side, you can have any number of products, but only a handful are "top level" products, the rest use a `/` e.g http://obofoundry.org/ontology/ro.html - note that base is in a subpath.

Historically I have just listed these top level ones in the products field on the purl yml, and used either partial or exact redirect rules for anything with a `/`. It may be cleaner to always list them explicitly in the yml products, but even so I feel uneasy about having too many things in the top level.

matentzn:
Not sure I follow this discussion: apart from adding trailing slashes to the urls under /about/ and /tracker/, what else do we need to do? Remove all the products apart from .obo and .owl and move all the secondary release products like gz etc under a bespoke releases folder?

PR #478, opened 2019-01-10
Update eupath.yml

johnwjudkins:
Updating for new release


jamesaoverton:
Is there a reason to remove the PURL for `/2018-05-09/obib.owl`? PURLs are supposed to be permanent, although in this case, it looks like it might have been an error.

johnwjudkins:
Yes, it is supposed to be /eupath.owl, not /obib.owl – we’re not removing the PURL, just correcting it.

Issue #251, opened 2016-10-17
Can we get rid of dependencies on OCLC for fallthru?


cmungall:
It turns out we still rely on OCLC for fallthrus

```
$ wget http://purl.obolibrary.org/obo/foobar
--2016-10-17 09:18:34--  http://purl.obolibrary.org/obo/foobar
Resolving purl.obolibrary.org (purl.obolibrary.org)... 52.3.123.63
Connecting to purl.obolibrary.org (purl.obolibrary.org)|52.3.123.63|:80... connected.
HTTP request sent, awaiting response... 302 Found
Location: http://purl.oclc.org/obo/foobar [following]
--2016-10-17 09:18:34--  http://purl.oclc.org/obo/foobar
Resolving purl.oclc.org (purl.oclc.org)... 207.241.224.2
Connecting to purl.oclc.org (purl.oclc.org)|207.241.224.2|:80... connected.
HTTP request sent, awaiting response... 302 FOUND
Location: http://www.berkeleybop.org/ontologies/foobar [following]
--2016-10-17 09:18:35--  http://www.berkeleybop.org/ontologies/foobar
```

we want the fallthrough to go direct to the obo-central server.

Yes, we should avoid fallthroughs and register all purls, even if explicitly to obo-central. But there are still hidden dependencies. This caused the GO pipeline to break at a bad point last week. cc @mcourtot @kltm

mcourtot:
Is the fall through due to the CNAME resolution from http://purl.obolibrary.org to occult or is it configured somewhere else?

jamesaoverton:
It's configured here: https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/www/.htaccess

It's very easy to change. I don't have any objections to changing it, just general anxiety. :)

mcourtot:
well on the bright side it can't be more broken than currently :)
I guess the correct link would be http://www.berkeleybop.org/ontologies/$1. @cmungall?

jamesaoverton:
This is still a problem, and maybe a bigger problem now than before.

I propose dropping the OCLC fall-through and serving our own 404 pages with some helpful documentation. What are the arguments **against** that proposal?

matentzn:
I agree with that.. dropping them.

PR #479, opened 2019-01-11
Update duo.yml

mcourtot:
Added home
Updated releases and about link to point to the releases page on GH and to the OLS

@jamesaoverton the fall through doesn't seem to work properly can you help? I didn't update this and was hoping that using http://purl.obolibrary.org/obo/duo/ would redirect to the GH repo by default but it instead points to berkeley

### DIFF
--- a/config/duo.yml
+++ b/config/duo.yml
@@ -17,13 +17,13 @@ entries:
   replacement: https://github.com/EBISPOT/DUO
   
 - prefix: /releases/
-  replacement: https://github.com/EBISPOT/DUO/releases/
+  replacement: https://raw.githubusercontent.com/EBISPOT/DUO/v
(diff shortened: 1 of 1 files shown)


jamesaoverton:
I'm seeing <http://purl.obolibrary.org/obo/duo/> redirect to <https://raw.githubusercontent.com/EBISPOT/DUO/master/>. I'm not sure what you want it to do.

If you drop the trailing slash, then <http://purl.obolibrary.org/obo/duo> goes to <http://ontologies.berkeleybop.org/>, but that's a top-level PURL that you're not supposed to use, and it's different issue about top-level failovers #251.

This PR looks fine, but doesn't change that `prefix: /` rule.

PR #483, opened 2019-01-28
Rename dpo.md to dpo.yml

matentzn:


matentzn:
I removed the example term because the check failed on travis.

PR #490, opened 2019-02-21
Update bfo.yml

zhengj2007:
Make a new BFO 2.0 (classes only) release and add BFO classes only development version

### DIFF
--- a/config/bfo.yml
+++ b/config/bfo.yml
@@ -4,7 +4,7 @@ idspace: BFO
 base_url: /obo/bfo
 
 products:
-- bfo.owl: https://raw.githubusercontent.com/BFO-ontology/BFO/v2.0/bfo.owl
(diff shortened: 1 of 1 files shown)


config/bfo.yml
```diff
@@ -132,9 +135,12 @@ entries:
 - exact: /2014-05-03/bfo.owl
   replacement: https://github.com/BFO-ontology/BFO/raw/releases/2014-05-03/owl-group/bfo.owl
 
-- exact: /2015-8-29/bfo.owl
```
jamesaoverton: This looks like it was a mistake, but I don't want to remove this PURL. I think we should keep `2015-8-29` and add `2015-08-29`.

zhengj2007: Actually this IRI do not implemented in the previous BFO OWL release file and used in anywhere. That's why I changed the link since it won't hurt anything. But it is a dated version IRI that pointed to first official release of BFO 2.0 (classes only) in case anyone would like to use in the future.

PR #493, opened 2019-02-27
Update zp.yml

matentzn:
We needed to switch to a different deposit system (zenodo) because of the sheer size of merged zp.owl. @cmungall for now we will do this manually. Is this the right way?


cmungall:
@matentzn is this ready to be merged?

matentzn:
will tell you next week, on holidays now.

Issue #63, opened 2015-11-18
Logs and Analysis


jamesaoverton:
This new system presents an opportunity to log and analyze OBO traffic. 
1. We should log failures to look for gaps in the new configuration, and other problems and strange behaviour.
2. We can provide metrics on OBO usage that could be very valuable, e.g. for demonstrating the use of our ontologies.

My experience, though, is that log analysis systems are either a real pain to set up or kind of expensive (or both). We don't have a lot of resources, time or money, for this.

I've built [ELK stacks](https://www.elastic.co/products) (Elasticsearch, Logstash, and Kibana) from open source parts. I found it a pain to set up, fragile, and pretty heavy: at least a medium sized server is required.

I don't have experience with commercial offerings, but there are many options, including hosted ELK:
- http://www.splunk.com
- https://papertrailapp.com
- http://aws.amazon.com/cloudwatch/

Suggestions? @cmungall @alanruttenberg @kltm

cmungall:
I'm not familiar with the different systems. I think this is v important though. If you and @kltm scope out the ideal system we can look at finding the resources to do this, possibly at OHSU

jamesaoverton:
I don't have any time left for this.

It doesn't have to get done tomorrow -- we'll just store our Apache logs for now. But I think it's important to the wider community, and it would be good to do it soon. I'll send an email about this to OFOC.

jamesaoverton:
Some thoughts on user privacy:

The system I have in mind would only analyze our Apache web server logs. It would not use cookies, and there would not be much user-specific information: just an IP address, the browser's UserAgent string, and the users HTTP Referrer (if provided). That's information that the user has explicitly sent to us in their HTTP request. It's helpful for counting "unique" users, and doing some crude geolocation, but not much else. We would not be sharing this information or trying to track users across multiple sites.

alanruttenberg:
I'll do some poking around. I'm in agreement that this is a great opportunity, as long as privacy and respect for our clients is forefront. I haven't looked at the logs, but maybe some simple scripting to analyze them may be useful. I think logs are a pain to analyze because of volume, but we may not have so much that we can't handle it with common programming.

jamesaoverton:
Even without a fancy logging system, I can make a Google Spreadsheet with a selection of failing paths, and we can discuss them on the OFOC call and obo-discuss list.

jamesaoverton:
[...]

PR #496, opened 2019-03-08
Initial port of genenontology/operations logrotate to S3 role to OBO

kltm:
This PR covers step "2" at https://github.com/OBOFoundry/purl.obolibrary.org/issues/63#issuecomment-457365282
Still requires a target bucket and S3cmd config (or other method of push, like awscli).

### DIFF
--- a/tools/server-overlay.yml
+++ b/tools/server-overlay.yml
@@ -1,0 +1,15 @@
+####
+#### WARNING: WIP
+#### This playbook is for ??.??.
+####  * Add remote logging to an already established server on S3.
+####
+---
+- hosts: all
+  remote_user: ubuntu
+  become: yes
+  roles:
+    - role: logrotate-to-s3
+      var_s3_path: "obo/purl"
+      var_name_prefix: ""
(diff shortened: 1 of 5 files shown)


tools/server-overlay.yml
```diff
@@ -0,0 +1,15 @@
...
+####
+---
+- hosts: all
+  remote_user: ubuntu
+  become: yes
+  roles:
+    - role: logrotate-to-s3
+      var_s3_path: "obo/purl"
+      var_name_prefix: ""
+      var_s3_push_credentials: "/home/sjcarbon/local/share/secrets/bbop/aws/s3/.s3cfg.obo-service-log-push"
```
kltm: This would need to be changed to the s3cfg file on your local system.

kltm: Tag onto #63

kltm: See: https://s3tools.org/kb/item14.htm for an example.
As well, I could send some examples, if useful.

jamesaoverton:
@cmungall said he would take care of this.

PR #505, opened 2019-04-09
Update geo.yml

dillerm:
Adding owl file purls for admin levels of Canada, China, and the UK.


cmungall:
FYI you can use the `prefix` tag for a generic redirect

PR #520, opened 2019-05-17
Add json-artefact

drseb:
HPO has added the JSON version of HP to the set of artefacts. It would be great to resolve this correctly via the purl "http://purl.obolibrary.org/obo/hp.json"


drseb:
ping @matentzn

PR #521, opened 2019-05-23
Update pr.yml (switched tracker to github)

nataled:


jamesaoverton:
@nataled I think that <https://github.com/PROconsortium/PRoteinOntology/issues> is a somewhat better URL to use, since it brings you to the list of all issues and lets you open a new one.

nataled:
True. I'll change it. I was used to how the SourceForge tracker worked, so set this up the same way. I am setting up some templates, and thought the user should be taken directly to the list of templates, but I see that clicking New Issue does indeed take you to the list. Thanks!

PR #527, opened 2019-06-11
Update iao.yml

zhengj2007:


config/iao.yml
```diff
@@ -10,30 +10,35 @@ base_url: /obo/iao
...
 example_terms:
 - IAO_0000030
 
+
 entries:
 
-- exact: /iao-stated.owl
-  replacement: https://raw.githubusercontent.com/information-artifact-ontology/IAO/master/releases/2017-03-24/iao.owl
+#- exact: /iao-stated.owl
+#  replacement: https://raw.githubusercontent.com/information-artifact-ontology/IAO/master/releases/2017-03-24/iao.owl
```
jamesaoverton: We should not drop PURLs if we can help it. I don't know why this PURL as made, but it should probably point to `iao.owl` in the latest release.

config/iao.yml
```diff
@@ -10,30 +10,35 @@ base_url: /obo/iao
...
 #- exact: /ontology-metadata.owl
 #  replacement: https://raw.githubusercontent.com/information-artifact-ontology/IAO/master/releases/2017-03-24/ontology-metadata.owl
-- exact: /iao-main.owl
-  replacement: https://raw.githubusercontent.com/information-artifact-ontology/IAO/master/releases/2017-03-24/iao-main.owl
+#- exact: /iao-main.owl
+#  replacement: https://raw.githubusercontent.com/information-artifact-ontology/IAO/master/releases/2017-03-24/iao-main.owl
```
jamesaoverton: This should probably point to the `iao-edit.owl` in the latest release.

config/iao.yml
```diff
@@ -68,8 +73,18 @@ entries:
 - prefix: /dev/
   replacement: https://raw.githubusercontent.com/information-artifact-ontology/IAO/master/src/ontology/
 
+- exact: /dev/iao-edit.owl
+  replacement: https://raw.githubusercontent.com/information-artifact-ontology/IAO/master/src/ontology/iao-edit.owl
```
jamesaoverton: The previous `prefix` entry covers this, so this `exact` entry will never be used.

PR #532, opened 2019-06-20
Create ornaseq.yml

zhengj2007:
issue: https://github.com/OBOFoundry/OBOFoundry.github.io/issues/618#

### DIFF
--- a/config/ornaseq.yml
+++ b/config/ornaseq.yml
@@ -4,9 +4,17 @@ idspace: ORNASEQ
 base_url: /obo/ornaseq
 
 products:
-- ornaseq.owl: https://raw.githubusercontent.com/safisher/ornaseq/master/ornaseq.owl
+- ornaseq.owl: https://raw.githubusercontent.com/safisher/ornaseq/v1.0/ornaseq.owl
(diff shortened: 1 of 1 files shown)


jamesaoverton:
This minimal PURL config is fine, and can be merged after https://github.com/OBOFoundry/OBOFoundry.github.io/pull/895

@safisher: The `ornaseq.owl` file does not have a VersionIRI. [OBO Principle 4: Versioning](http://www.obofoundry.org/principles/fp-004-versioning.html) and our best practises (and the FAIR principles, more generally) require release artifacts (files) to specify versions, so we can continue to refer to old versions. Once the OWL file has a VersionIRI, there should be a PURL pointing to that version.

jamesaoverton:
As pointed out here https://github.com/OBOFoundry/OBOFoundry.github.io/issues/618#issuecomment-509353852 there is now a VersionIRI: http://purl.obolibrary.org/obo/ornaseq/2019-07-08/ornaseq.owl

The `ornaseq.yml` needs an entry pointing to that VersionIRI, like this https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/opl.yml#L14

PR #533, opened 2019-06-26
Hardcode redirect for broken version IRI

balhoff:
See https://github.com/geneontology/pipeline/issues/90


balhoff:
cc @cmungall @kltm

Issue #534, opened 2019-07-07
Provide an easier way to use a custom browser for class PURLs


cmungall:
We have a growing number of opt-outs of ontobee in obo.yml. This is only likely to grow.

Ideally we would not have lots of people editing obo.yaml

It's not clear to me if it's actually necessary to edit obo.yml to redirect to OLS? I will try an experiment and link here.

PR #535, opened 2019-07-07
WIP: Experimenting with OLS instead of ontobee for ceph

cmungall:
See #534

### DIFF
--- a/config/ceph.yml
+++ b/config/ceph.yml
@@ -13,7 +13,7 @@ example_terms:
 
 entries:
 - prefix: /about/
-  replacement: https://www.ebi.ac.uk/ols/ontologies/ceph/terms?iri=http://purl.obolibrary.org/obo/CEPH_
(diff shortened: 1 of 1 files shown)


config/ceph.yml
```diff
@@ -7,11 +7,16 @@ products:
...
 
-term_browser: ontobee
+term_browser: custom
 example_terms:
-- CEPH_0000000
+- CEPH_0000256
 
 entries:
+- prefix: /about/
+  replacement: https://www.ebi.ac.uk/ols/ontologies/ceph/terms?iri=http://purl.obolibrary.org/obo/CEPH_
```
jamesaoverton: Now that the `tools/test.py` is fixed, this fails because there ends up being an extra `CEPH_`.

jamesaoverton:
@cmungall Is that what you wanted? `term_browser: ols` and `/about/*` going to OLS?

jamesaoverton:
I guess this is not a priority. Feel free to reopen.

PR #536, opened 2019-07-07
Add OLS option for term_browser, add poro as test case

### LINKED ISSUE #534: Provide an easier way to use a custom browser for class PURLs

cmungall:
Fixes #534

### DIFF
--- a/config/poro.yml
+++ b/config/poro.yml
@@ -9,9 +9,9 @@ products:
 - poro.owl: https://raw.githubusercontent.com/obophenotype/porifera-ontology/master/poro.owl
 - poro.obo: https://raw.githubusercontent.com/obophenotype/porifera-ontology/master/poro.obo
 
-term_browser: ontobee
+term_browser: ols
 example_terms:
-- PORO_0000000
+- PORO_0000017
 
 entries:
 - exact: /browse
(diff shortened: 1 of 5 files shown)


jamesaoverton:
This looks good. Is it ready to merge?

cmungall:
I didn't get round to testing this on my machine

config/poro.yml
```diff
@@ -23,8 +23,8 @@ entries:
 - prefix: /about/
   replacement: http://www.ontobee.org/browser/rdf.php?o=PORO&iri=http://purl.obolibrary.org/obo/
```
jamesaoverton: This replacement still points to Ontobee, so the test that points to EBI fails (and with #544 we can now see the failure).

The `term_browser` field is specifically for the `http://purl.obolibrary.org/obo/ABC_123` case. There's no special handling of `/about/`.

jamesaoverton:
Now I see that one of the PORO test is failing. The script changes adding `ols` ares fine. I want to add `ols` to `tools/test.py` as well.

jamesaoverton:
@simonjupp For the term "PORO:0000017" the PURL is <http://purl.obolibrary.org/obo/PORO_0000017> and with this PR we'll redirect it to OLS at <https://www.ebi.ac.uk/ols/ontologies/poro/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FPORO_0000017>. Is this right? In particular:

1. should we be lowercasing our `idspace: PORO` to `poro` in the middle of the URL?
2. should we be escaping slashes and stuff in the query part of the URL?

matentzn:
@henrietteharmse @udp Can you guys help @jamesaoverton  out here? Thanks!

udp:
Hello,

1. We use the IDs from https://raw.githubusercontent.com/OBOFoundry/OBOFoundry.github.io/master/_config.yml which are indeed lowercase (poro)

2. Yes, these characters need to be escaped as they are part of the query string of the URL.

Let me know if you have any more questions.

jamesaoverton:
Thanks @udp and @matentzn!

PR #538, opened 2019-07-10
Update base_redirect for DO

rctauber:
Allow navigation to other products (without listing them all as `entries`)

### DIFF
--- a/config/doid.yml
+++ b/config/doid.yml
@@ -3,7 +3,7 @@
 idspace: DOID
 base_url: /obo/doid
 
-base_redirect: https://github.com/DiseaseOntology/HumanDiseaseOntology/tree/master/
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Can `HumanDO.obo` be pointed to something, even if it's not current?

rctauber:
With the new base, `http://purl.obolibrary.org/obo/doid/HumanDO.obo` should still point to https://raw.githubusercontent.com/DiseaseOntology/HumanDiseaseOntology/master/src/ontology/HumanDO.obo

`doid-non-classified.obo` and `HumanDO.obo` are exactly the same, just with different names (https://github.com/DiseaseOntology/HumanDiseaseOntology/blob/master/Makefile#L151)

PR #544, opened 2019-07-22
test.py must exit 1 on any failure

jamesaoverton:


jamesaoverton:
The `tools/test.py` script has not been exiting with a non-zero code when failures are found in the HTTP tests. This must go back to the recent refactoring, and I missed the problem in my review. I only noticed the problem today when reviewing #535.

This PR makes the script exit 1 on any failure, which Travis will then recognize as a failure. The only problem that I found was with #537 which I fixed here.

PR #547, opened 2019-07-25
Update dpo.yml

matentzn:
DPO has a new home now


matentzn:
@jamesaoverton I cant see why this fails :/ any idea? Remember this is one of these weird cases where dpo is the ontology id, but the idspace is actually FBcv.. But given that I do not change anything like that..

matentzn:
Ok it failed because of the example term provided (because it was in the wrong namespace). Fixed now!

PR #548, opened 2019-07-26
Update fbcv.yml

matentzn:
New release structure; added redirects to old releases.

Please dont merge until @dosumis checks this.

@jamesaoverton do you think the redirects work like that? Previously, there was a 'releases' folder on master; now releases should be redirected to the usual Github releases (reflected by the default /releases/ redirect).


matentzn:
I will wait for James to take a quick peak as well; Thanks!

matentzn:
Thanks both!

PR #551, opened 2019-08-02
Update zeco.yml

matentzn:
Enable generic fall-through redirect.

### DIFF
--- a/config/zeco.yml
+++ b/config/zeco.yml
@@ -6,6 +6,7 @@ base_url: /obo/zeco
 products:
 - zeco.owl: https://raw.githubusercontent.com/ybradford/zebrafish-experimental-conditions-ontology/master/zeco.owl
 - zeco.obo: https://raw.githubusercontent.com/ybradford/zebrafish-experimental-conditions-ontology/master/zeco.obo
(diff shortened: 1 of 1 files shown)


althonos:
@matentzn : seems like you didn't rebuild the site, `zeco.md` still contains `zeco.obo` twice! http://www.obofoundry.org/ontology/zeco.html

matentzn:
@althonos I cant do that; This pull request only made sure that

http://purl.obolibrary.org/obo/zeco.json

resolves correctly!

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/ontology/zeco.md

determines what products are shown on the site. Changes on the life site only apply once Chris re-deploys the site at some point!

althonos:
Bump @cmungall then :grin:

jamesaoverton:
A change was required to [zeco.md](https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/ontology/zeco.md), which I made in https://github.com/OBOFoundry/OBOFoundry.github.io/commit/c8d053bbe411479f643e647bba946730bd93e327. So this page should now be correct: http://www.obofoundry.org/ontology/zeco.html

matentzn:
James life safer :P Ahhhh I am too hasty!

PR #559, opened 2019-08-09
Corrections to ORNASEQ yml file

safisher:
I believe I've now correctly updated the current release of the ontology and this yml update should point the PURL to the appropriate owl/obo files.

### DIFF
--- a/config/ornaseq.yml
+++ b/config/ornaseq.yml
@@ -4,8 +4,8 @@ idspace: ORNASEQ
 base_url: /obo/ornaseq
 
 products:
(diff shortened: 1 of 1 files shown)


jamesaoverton:
PURLs are supposed to be permanent, but `/releases/2018-06-04/ornaseq.owl` and `.obo` were being removed by this PR. I made an edit that restores those PURLs to point to the v1.0 release, as they did in the previous version of this config file. Do you agree?

The other changes are fine.

safisher:
@jamesaoverton you are correct, I incorrectly made that change. I also made another mistake with the product URL, that I've now fixed.

PR #560, opened 2019-08-13
Update mp.yml

matentzn:
Hopefully the fall through works like that.. :)


jamesaoverton:
Sorry, I was away for a few days.

I think this is what you want. Does `/mp-base.owl` still point to something?

matentzn:
No worries. This is now caught by the fall through!

PR #562, opened 2019-08-13
Update doid.yml

beckyjackson:
Fix DOID redirects.

### DIFF
--- a/config/doid.yml
+++ b/config/doid.yml
@@ -3,11 +3,12 @@
 idspace: DOID
 base_url: /obo/doid
 
-base_redirect: https://github.com/DiseaseOntology/HumanDiseaseOntology/tree/master/src/ontology/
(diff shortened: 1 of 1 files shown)


config/doid.yml
```diff
@@ -3,26 +3,33 @@
...
 - doid.obo: https://raw.githubusercontent.com/DiseaseOntology/HumanDiseaseOntology/master/src/ontology/doid.obo
+- doid.json: https://raw.githubusercontent.com/DiseaseOntology/HumanDiseaseOntology/master/src/ontology/doid.json
 
 term_browser: ontobee
 example_terms:
 - DOID_4
 
 entries:
 
+- exact: /
```
jamesaoverton: `exact` can't have tests. Should this be `prefix`? If so, it should be the last entry, not the first, since entries are checked in the order they appear.

beckyjackson: Whoops, yes, this should be `prefix`. Fixed.

PR #565, opened 2019-08-16
Create sepio.yml

matentzn:
There was for a very long time an [md file](https://github.com/OBOFoundry/OBOFoundry.github.io/edit/master/ontology/sepio.md), but no corresponding PURL config..

### DIFF
--- a/config/sepio.yml
+++ b/config/sepio.yml
@@ -1,0 +1,26 @@
+# PURL configuration for http://purl.obolibrary.org/obo/sepio
+
+idspace: SEPIO
+base_url: /obo/sepio
+
+products:
+- sepio.owl: https://github.com/monarch-initiative/SEPIO-ontology/releases/download/current/sepio.owl
+
+term_browser: ontobee
+example_terms:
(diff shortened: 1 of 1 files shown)


config/sepio.yml
```diff
@@ -0,0 +1,26 @@
...
+
+- prefix: /tracker/
+  replacement: https://github.com/monarch-initiative/SEPIO-ontology/issues
+
+- prefix: /about/
+  replacement: http://www.ontobee.org/ontology/SEPIO?iri=http://purl.obolibrary.org/obo/
+
+## generic fall-through, serve direct from github by default
+- prefix: /
+  replacement: https://github.com/monarch-initiative/SEPIO-ontology/releases/download/current/
```
jamesaoverton: Most projects fall-through to `master`, but falling-through to the current release is also fine.

matentzn: Yeah, there is currently nothing on master, that’s why I redirect here for now. I will change if the situation changes!

PR #567, opened 2019-08-26
Update bfo.yml

zhengj2007:


zhengj2007:
Add new release which made minor changes on the ontology information, details see: https://github.com/BFO-ontology/BFO/releases/tag/v2019-08-26

PR #568, opened 2019-08-29
Create labo.yml

pfabry:
Adding yml file for LABO ontology.

### DIFF
--- a/config/labo.yml
+++ b/config/labo.yml
@@ -1,0 +1,11 @@
+# PURL configuration for http://purl.obolibrary.org/obo/labo
+
+idspace: LABO
+base_url: /obo/labo
+
+products:
+- labo.owl: https://github.com/OpenLHS/LABO/blob/master/LABO.owl
+
+term_browser: ontobee
+example_terms:
+- LABO_0000007
--- a/config/labo.yml
+++ b/config/labo.yml
@@ -4,7 +4,7 @@ idspace: LABO
 base_url: /obo/labo
 
 products:
-- labo.owl: https://github.com/OpenLHS/LABO/blob/master/LABO.owl
+- labo.owl: https://raw.githubusercontent.com/OpenLHS/LABO/master/Releases/2019-08-29/LABO.owl
 
 term_browser: ontobee
 example_terms:


jamesaoverton:
Thanks for the PR. There are two problems that have to be fixed before we can merge:

1. The `labo.owl` PURL must point to a file in RDF/XML format (see [Principle 2](http://www.obofoundry.org/principles/fp-002-format.html)). It currently points to a GitHub HTML page for a file. That could be fixed by pointing to the "raw" file at https://raw.githubusercontent.com/OpenLHS/LABO/master/LABO.owl. However that file is in OWL Functional syntax. It's better for the published versions of the ontology to be in RDF/XML, which can be read by general RDF tools and not just OWL tools.

2. It's also important for the published file to have a version IRI that will be maintained with a PURL. See [Principle 4](http://www.obofoundry.org/principles/fp-004-versioning.html).

pfabry:
Hello James,

Thank you for the feedback. I just did the following:
 - added a version IRI;
 - RDF/XML output format;
 - modified the labo.yml with the new (and correct, sorry for that…) link;
 - created a labo.md file in OBOFoundry.github.io/ontology/<http://OBOFoundry.github.io/ontology/>

I hope this will meet your requirements.

Paul

Le 29 août 2019 à 10:47, James A. Overton <notifications@github.com<mailto:notifications@github.com>> a écrit :

Thanks for the PR. There are two problems that have to be fixed before we can merge:

  1.  The labo.owl PURL must point to a file in RDF/XML format (see Principle 2<http://www.obofoundry.org/principles/fp-002-format.html>). It currently points to a GitHub HTML page for a file. That could be fixed by pointing to the "raw" file at https://raw.githubusercontent.com/OpenLHS/LABO/master/LABO.owl. However that file is in OWL Functional syntax. It's better for the published versions of the ontology to be in RDF/XML, which can be read by general RDF tools and not just OWL tools.

  2.  It's also important for the published file to have a version IRI that will be maintained with a PURL. See Principle 4<http://www.obofoundry.org/principles/fp-004-versioning.html>.

jamesaoverton:
I'll merge this, but I have two suggestions:

1. You should include your CC-BY 4.0 license in the published OWL file
2. I recommend using the GitHub Releases mechanism for publishing releases, rather than just a folder in the repository

PR #584, opened 2019-09-26
Update pr.yml

nataled:


jamesaoverton:
I'm not sure what the intention is here, but PURLs should not be removed, and this seems to be dropping `/pr.owl`

nataled:
So far as I can tell, all PURLS are now working as intended. The drops I did were to correct the new ones that weren't quite correct. They were only in place for no more than 30 minutes or so, and not advertised.

jamesaoverton:
Ok, that makes sense. Thanks.

PR #586, opened 2019-10-10
Update fbbt.yml

matentzn:
Oversight when updating urls..


jamesaoverton:
There was a sudden, unrelated change to Python dependencies that broke this build. Your changes are fine. I'll just merge.

PR #587, opened 2019-10-11
Switch all CHEBI URLs to HTTP

matentzn:
Travis [does not play well with FTP](https://docs.travis-ci.com/user/common-build-problems/#ftpsmtpother-protocol-do-not-work), so this just needed happening.

### DIFF
--- a/config/chebi.yml
+++ b/config/chebi.yml
@@ -4,10 +4,10 @@ idspace: CHEBI
 base_url: /obo/chebi
 
 products:
-- chebi.owl: ftp://ftp.ebi.ac.uk/pub/databases/chebi/ontology/chebi.owl
-- chebi.obo: ftp://ftp.ebi.ac.uk/pub/databases/chebi/ontology/chebi.obo
-- chebi.owl.gz: ftp://ftp.ebi.ac.uk/pub/databases/chebi/ontology/chebi.owl.gz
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Hmm. We’ve been using FTP for 2.5 years or so. I’ve also run into some problems, but worked around them. Does everyone agree that switching to HTTP is a good idea?

matentzn:
Your Travis jobs pulling chebi, do any of them work?

jamesaoverton:
I guess I don’t have any.

Fetching with what? `robot`? `curl`?

matentzn:
I think the problem is independent of the method but not sure. I tried robot and wget. Switching to http solved the issue!

matentzn:
What are your worries switching to http?

jamesaoverton:
Do you think Travis is just blocking FTP ports?

I'm worried that people have already written custom code to handle ChEBI FTP as a special case, and this will break their workflows. I'm not vetoing this change. I just want to hear a second opinion.

@cmungall What do you think?

cmungall:
If people have already hard coded the ftp URLs they won't be affected. I
don't see any downsides or what can go wrong. And this will solve issues. I
been burned by weird behavior with ftp and owl api before

jamesaoverton:
Ok. @matentzn please go ahead and merge.

matentzn:
Thanks for the input! :)

PR #590, opened 2019-10-29
Fix the PURL for /2010-05-25/ to point to raw

alanruttenberg:
and include trailing "/"


alanruttenberg:
Was missing trailing "/" and didn't point to github raw

PR #594, opened 2019-11-12
Changed location of main ZP release to github

matentzn:


jamesaoverton:
Out of curiosity, is there an advantage to `/releases/downloads` over `raw.githubusercontent.com`?

matentzn:
Actually, no idea! I was using this for [MP](https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/mp.yml) for a while now, and it seems to work.. No idea.. Sorry!

PR #596, opened 2019-11-18
Update bto.yml

matentzn:
Redirect to new location

### DIFF
--- a/config/bto.yml
+++ b/config/bto.yml
@@ -15,7 +15,7 @@ example_terms:
 entries:
 
 - prefix: /releases/
-  replacement: https://github.com/BRENDA-Enzymes/BTO/v
+  replacement: https://raw.githubusercontent.com/BRENDA-Enzymes/BTO/v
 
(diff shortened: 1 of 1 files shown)


config/bto.yml
```diff
@@ -4,8 +4,25 @@ idspace: BTO
...
+- bto.json: https://github.com/BRENDA-Enzymes/BTO/raw/master/bto.json
 
 term_browser: ontobee
 example_terms:
 - BTO_0000000
+
+entries:
+
+- prefix: /releases/
+  replacement: https://github.com/BRENDA-Enzymes/BTO/v
```
jamesaoverton: I was expecting this to point to a raw file, not HTML.

matentzn: An that is why i still need people looking over my shoulder. Thanks for the diligence @jamesaoverton should be fixed now!

jamesaoverton: No problem. I approve this. Please merge when you're ready.

PR #598, opened 2019-11-20
Update OBI config for release 2019-11-12

beckyjackson:


config/obi.yml
```diff
@@ -33,9 +33,6 @@ entries:
 - exact: /obi_FGED.owl
   replacement: http://svn.code.sf.net/p/obi/code/releases/2013-10-25/obi_FGED.owl
 
-- exact: /obi.obo
```
jamesaoverton: This PURL points to `/obo/obi/obi.obo`, which is not the same as `/obo/obi.obo` in the products. We can't drop it.

beckyjackson: Fixed

PR #606, opened 2020-02-06
DO NOT MERGE (yet): Update uberon.yml

balhoff:
See https://github.com/obophenotype/uberon/issues/1559


matentzn:
I just ran wget 5 times on bbop vs svn and svn seems more about 4 times slower.. Can you check this as well?

balhoff:
I agree: a few seconds for SF, negligible time for berkeleybop.

matentzn:
Superseded by #731

PR #610, opened 2020-02-14
Update pco.yml

ramonawalls:
Adding replacement for /pco_merged_inferred.owl back in. Even though the file is not kept up to date anymore, the PURL should be permanent. Thanks @jamesaoverton for the reminder.


jamesaoverton:
Thanks @ramonawalls !

PR #619, opened 2020-03-12
Create cido.yml

yongqunh:
CIDO has been discussed and approved. 
Reference: https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1123


yongqunh:
Thanks, James!

PR #623, opened 2020-03-20
Fixes release redirects for CDAO to use regex

hlapp:
This is so that not each and every single release (or release year) needs to be added. It turns out the June 2019 release was missed to be added here, rendering the version IRI non-resolving.

### DIFF
--- a/config/cdao.yml
+++ b/config/cdao.yml
@@ -23,19 +23,17 @@ entries:
   - from: /1.0/cdao.owl
     to: https://raw.githubusercontent.com/evoinfo/cdao/cdao-rel-1.0/OWL/cdao.owl
 
-- prefix: /2012-
-  replacement: https://raw.githubusercontent.com/evoinfo/cdao/cdao-rel-2012-
+- regex: ^/obo/cdao/(\d+-\d+-\d+)/(\S+)$
+  replacement: https://raw.githubusercontent.com/evoinfo/cdao/cdao-rel-$1/$2
   tests:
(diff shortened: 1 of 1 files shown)


jamesaoverton:
I'm happy to merge this, but note that while `prefix:` builds case-insensitive regular expressions (see [the code](https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/tools/translate_yaml.py#L175) and [the README](https://github.com/OBOFoundry/purl.obolibrary.org#regex)), `regex:` uses exactly what you specify.

I tend to prefer `prefix: /20` over a regex for these case ([example](https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/mro.yml#L16)) but that's debatable.

If you want exactly this, that's fine I'll merge, just please confirm.

hlapp:
I was indeed debating myself between these alternatives. Here (as I think in most other OBO ontologies) only the OWL ontology itself is versioned; the term IRIs are not, so while it looks like it's meant to capture a lot, in reality this will only be touched for the ontology and ontology version IRIs.

Is the concern here that case-sensitivity would also apply to the domain name? The code doesn't look like it would (it doesn't seem to prefix the regex with the domain). So provided I'm right with my understanding that only the path (`/obo/cdao/`) would be case sensitive, I'm fully OK with that.

jamesaoverton:
Thanks for the reply @hlapp. Yes, the system only looks at the path part of the IRI for matching, not the domain name.

PR #626, opened 2020-03-27
Update gno.yml

edwardsnj:


jamesaoverton:
Hey @edwardsnj. I got distracted and clicked "merge" too quickly. I have a question about this PR:

Our PURLs should be Permanent, so you should never delete a PURL entry, just add new ones. I'm going to restore the one that you deleted in this PR.

If that's a problem, let me know.

PR #630, opened 2020-03-27
Update bco.yml

ramonawalls:
Update file so that it now points to the bco.owl file in the repo root directory as the main release file, consistent with ODK repo structuring.

Add new prefix for this release.

No longer releasing a "non-classified" version with this release, but it may be added back in with the next release.

### DIFF
--- a/config/bco.yml
+++ b/config/bco.yml
@@ -4,7 +4,7 @@ idspace: BCO
 base_url: /obo/bco
 
 products:
-- bco.owl: https://raw.githubusercontent.com/BiodiversityOntologies/bco/2016-12-14/bco.owl
+- bco.owl: https://raw.githubusercontent.com/BiodiversityOntologies/bco/master/bco.owl
 
 term_browser: ontobee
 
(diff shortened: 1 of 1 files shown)


jamesaoverton:
This change is fine, but I want to double-check something with you:

ODK doesn't actually require that the PURL for your primary product `bco.owl` points to the `master` branch -- you can point it to a dated release. People use `master` for their primary OWL and a `prefix:` pointing to GitHub releases so that they don't have to update this PURL config file on each new release. That's independent of ODK, but lots of ODK-users do it this way.

If you're happy with your change, I'll go ahead and merge it, just let me know.

ramonawalls:
Yes, I am fine with having it point to master. The current editing
structure supports it.

Thanks, James!
------------------------------------------------------
Ramona L. Walls, Ph.D.
Assistant Research Professor, Bio5 Institute, University of Arizona
Part-time artist (https://www.instagram.com/omnivoresgratitutude/)

PR #635, opened 2020-04-10
Updating ceph.yml to use OLS

cmungall:


jamesaoverton:
`term_browser: custom` requires a corresponding entry in `obo.yml`. DUO is an example that uses OLS: https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/obo.yml#L39

We have a WIP PR #536 to allow `term_browser: ols` which would be a better solution here.

jamesaoverton:
Closing this in favour of #535

PR #637, opened 2020-04-14
correcting github raw link format

Robbie1977:
http://purl.obolibrary.org/obo/so.owl is currently returning 404 as old raw URL is used.

### DIFF
--- a/config/so.yml
+++ b/config/so.yml
@@ -4,8 +4,8 @@ idspace: SO
 base_url: /obo/so
 
 products:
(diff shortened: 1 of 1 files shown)


jamesaoverton:
The problem with the old links is just that they were missing the `Ontology_Files/` part.

These new `github.com` links simply redirect to `raw.githubusercontent.com`. So I suggest just fixing your `raw.githubusercontent.com` URLs and avoiding the extra layer of redirection.

Robbie1977:
@jamesaoverton changed

jamesaoverton:
Thanks!

PR #651, opened 2020-05-03
Create omo.yml

zhengj2007:
Add file for newly registered OBO metadata ontology (OMO)

### DIFF
--- a/config/omo.yml
+++ b/config/omo.yml
@@ -1,0 +1,23 @@
+# PURL configuration for http://purl.obolibrary.org/obo/omo
+
+idspace: OMO
+base_url: /obo/omo
+
+products:
+- omo.owl: https://raw.githubusercontent.com/information-artifact-ontology/ontology-metadata/v2020-05-03/omo.owl
+
+term_browser: ontobee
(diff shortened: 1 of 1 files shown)


jamesaoverton:
If some IAO terms are going to redirect to OMO, then we need to carefully override the main term redirects like this https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/obo.yml#L22

If you give me a list of the overrides I can do that.

zhengj2007:
@jamesaoverton Here is the list of terms. Thanks!
[termlist.txt](https://github.com/OBOFoundry/purl.obolibrary.org/files/4581560/termlist.txt)

jamesaoverton:
#653 implements the required change. But in order for this to work, OMO needs to be on Ontobee.

jamesaoverton:
I merged this into #653, so I'll close this PR.

jamesaoverton:
After discussion in #653, let's move ahead with the original file, minus the IAO example term.

jamesaoverton:
@zhengj2007 Is this ready to go?

PR #653, opened 2020-05-05
Add adoptions.tsv to configure term overrides

jamesaoverton:
So far we've had four cases where we redirect a term to a different ontology than its IDSPACE: four BFO IDs that ended up in RO instead.

With the addition of OMO in #651 we need to override a list of IAO terms. So I created an `adoptions.tsv` table to contain all these mappings, and hacked the code to write redirects for each *before* the primary term ID redirects.

I think this is fine and just needs an addition to the README, but I'm going to sleep on it.


jamesaoverton:
Here's a longer discussion that I sent to the TWG mailing list:

I think this is the simplest and most direct solution to a problem that we've recognized for a while: migrating the responsibility for selected terms from one OBO project to another without changing the term ID. I'm calling this "term adoption".

We've had four key examples of term adoption our new PURL system was built in 2015: 'part of' has ID BFO:0000050 but RO has responsibility for that term, so we redirect <http://purl.obolibrary.org/obo/BFO_0000050> to RO's Ontobee page. That's configured here:

https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/obo.yml#L24

The new OBO Metadata Ontology project is taking responsibility for a number of IAO terms, so I built this table to make that same redirect process a little more convenient:

https://github.com/OBOFoundry/purl.obolibrary.org/blob/adoptions/config/adoptions.tsv

So if a term is not in this adoption table, we look at the IDSPACE for the term and redirect based on that.

Other than BFO/RO and IAO/OMO, we could also use this approach to solve longstanding problems with D-ACTs and Proper Names Ontology. I know of a handful of assays that OBI has adopted from CMO.

There are other ways we could do this. The main disadvantage that I see to this approach is that if users aren't going through our PURL system, they need to check our adoption table themselves.

The main alternative to this is: Don't. Just obsolete the term in IAO and mint a new ID in OMO. For the BFO/RO cases we decided that would cause too much churn. Likewise these IAO terms are used a *lot*. Maybe there's a line to be drawn.

cmungall:
I like the system

still think it should system of last resort

Do we really need this for OM? Thought we were going to mint new stuff in OM space, but have OM bring in things from outside (dct, iao, stable IAO terms)

jamesaoverton:
@cmungall I'm not sure I understand what you're asking. OMO is taking responsibility for IAO:0000115 'definition', used to annotate approximately every OBO term. What do you want the PURL system to do for IAO:0000115?

alanruttenberg:
Consider recording this information in the yaml config instead. e.g.

Adopts:
 - xxxx
 - yyyy

Puts control of this in the hands of the developers rather than the purl maintainer have to do something

If there's a concern that there might be adopting terms without the source's constant, then require that the original ontology declares which terms it donates:

Donates:
 - xxx
 - yyy

I'm mixed about whether Donates should also record the destination ontology, but lean to not.

Easy enough to check that there aren't conflicts when the testing is done.

zhengj2007:
For the terms in OMO, I'd prefer to keep IAO IDs since they have been widely used in OBO Foundry ontologies.

What Alan proposed to solve the ID overrides looks good to me if it is not too hard to achieve.

Regarding the D-ACTs and Proper Names Ontology, I'd prefer to merge them with IAO. We can keep them in separate OWL files and make D-ACTs, PNO release like GO slim/OBI view if need.

ramonawalls:
I really appreciate the solution, @jamesaoverton . I like that @alanruttenberg 's suggestion puts the burden on the developers, but then I have no idea how hard it would be to implement. I support keeping the original IRIs for widely used terms. For the ontologies I manage, it would not be too terribly difficult to replace them, but I don't know what the downstream effects would be.

cmungall:
@jamesaoverton - I'm suggesting that in the case of IAO:0000115 is just does the default redirect to IAO. OMO has not formally adopted the term. OMO is just extracting the relevant subset of IAO, just as it extracts subsets of other ontologies, as well as minting any new terms required in its own ID space.

There are downsides. It might be better for OMO to formally adopt, it can then do things like modify the definition of definition or add other metadata. At the moment formally it has to request any changes via IAO. But I expect the subset of IAO imported terms to be fairly stable.

jamesaoverton:
@alanruttenberg's suggestion makes adoptions more visible than a separate table, and isn't too much harder to implement. It might still be good to compile a table that downstream tools can use more easily than the YAML. I'll work on that. Thanks.

@cmungall I believe that @zhengj2007 has been working toward adoption as I have implemented it, and not what you describe. She has been pulling those IAO terms into OMO and then importing OMO into IAO:

https://github.com/information-artifact-ontology/IAO/blob/master/src/ontology/iao-edit.owl#L22

If we do not adopt, then a new IAO release will be required every time OMO is updated in order for Ontobee etc. to reflect the change to OMO. That isn't the end of the world, but I thought one of the goals was to decouple OMO and IAO releases.

I don't have strong opinions either way. I'm just trying to implement what I thought we had agreed on.

cmungall:
Ah, I see the OMO PURL request is bundled into this PR. Can we separate?

I think the OMO PURL request should go through ASAP

I also think the technical implementations should go through ASAP

What I would like more time to discuss is whether we need such a large scale adoption of IAO IDs. I think adoption should be a system of last resort. I believe that OMO can progress in the interim by simply importing an extract of IAO. It may be the case that this poses some problems downstream, but we can always adopt later, this will be transparent to users. This should be a separate decision that the decision to adopt the technical system (which is great) and the OMO PURL request (also great).

jamesaoverton:
@cmungall I think you're proposing that we merge an `omo.yml` without adoptions right now. The IAO terms in OMO will continue to resolve to IAO on Ontobee. Then we will discuss whether those terms should be formally adopted into OMO. Please correct me if I'm mischaracterizing.

@zhengj2007 Are you OK with that proposal?

cmungall:
> The IAO terms in OMO will continue to resolve to IAO on Ontobee. Then we will discuss whether those terms should be formally adopted into OMO. Please correct me if I'm mischaracterizing.

This is correct. IMO This is not so different than OMO blessing and importing, say, dcterm:foo. OMO is assembling terms from different sources as needed and minting new ones in the gaps.

We may formally adopt some later, but I don't think there is a rush, and I would like to do this in a measured way. For example, maybe 'universal' (http://purl.obolibrary.org/obo/IAO_0000410) would never be adopted by OMO.

zhengj2007:
> @cmungall I think you're proposing that we merge an `omo.yml` without adoptions right now. The IAO terms in OMO will continue to resolve to IAO on Ontobee. Then we will discuss whether those terms should be formally adopted into OMO. Please correct me if I'm mischaracterizing.
> 
> @zhengj2007 Are you OK with that proposal?

@jamesaoverton Yes. I am OK with it. I am planning to make an IAO release soon after the release of OMO. We can discuss the term overriding issue in the future.

PR #661, opened 2020-05-27
Creation of ons.yml

FrancescoVit:
See https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1198

### DIFF
--- a/config/ons.yml
+++ b/config/ons.yml
@@ -4,8 +4,8 @@ idspace: ONS
 base_url: /obo/ons
 
 products:
-- ons_release_v1.0.owl: https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/ons_release_v1.0.owl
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Hi @FrancescoVit. The `products` name should be simply your IDSPACE an the file extension, and should not contain any version information. I made that change. Is this acceptable?

It's good practise to include PURLs for versions as well, but this should be done in the `entries` section. For example:

```
entries:
  - exact: ons_release_v1.0.owl
     replacement: https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/ons_release_v1.0.owl
```

This will redirect <http://purl.obolibrary.org/obo/ons/ons_release_v1.0.owl> to <https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/ons_release_v1.0.owl>.

FrancescoVit:
Thank you @jamesaoverton , the edit is acceptable.

PR #662, opened 2020-06-04
Refactor PURL JSON schema to re-use definitions

jamesaoverton:
I asked @jannahastings to update the JSON Schema for the PURL system for better reuse, as described here:

https://json-schema.org/understanding-json-schema/structuring.html#reuse


cmungall:
Much prefer this style, thanks!

PR #669, opened 2020-06-27
change uberon to use OLS

cmungall:


jamesaoverton:
Do you want to use `term_browser: ols`?

PR #680, opened 2020-08-24
Update ontology version

FrancescoVit:


jamesaoverton:
Hi @FrancescoVit. Your changes to the `products` section are good. However it's important not to remove any PURL entries. PURLs should permanent, so that your users are able to access old versions of your ontology at stable URLs.

I tried reverting the deletion of the `ons_release_v1.0.owl` under `entries`, but then I realized that the `replacement` URL no longer resolves. So I looked around your GitHub repository and found the older file (I think) and updated the PURL entry.

Is this acceptable?

FrancescoVit:
Hi @jamesaoverton  and thank for the help. Yes, while updating to a new version, I kept track of older version in a dedicated folder. The update is acceptable, for future versioning I'll keep track of PURLs in the entries section.

PR #682, opened 2020-08-27
Update hp.yml

drseb:
- remove uberpheno link
- add hpoa files

ping @matentzn


matentzn (approved):
I have tested the links and the syntax of the entries seems correct. I will merge when you change the status from Draft to normal and QC has done its thing.

PR #685, opened 2020-09-11
Add IHCC view to GECKO products

beckyjackson:
Adds `ihcc-gecko.owl` to products. Fixing the IRI for https://github.com/IHCC-cohorts/GECKO/issues/2

Also adds a `/releases/` path.

### DIFF
--- a/config/gecko.yml
+++ b/config/gecko.yml
@@ -5,6 +5,7 @@ base_url: /obo/gecko
 
 products:
 - gecko.owl: https://raw.githubusercontent.com/IHCC-cohorts/GECKO/master/gecko.owl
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Since the project IDSPACE is `gecko`, all the products must be `gecko.*`. So this `ihcc-gecko.owl` product should fail the Travis tests. I'm not sure why it's passing.

Please:

1. Fix the tests to enforce the IDSPACE for `products` so that this PR fails.
2. Fix the `gecko.yml` by removing the `ihcc-gecko.owl` product and adding an entry like:

```
- exact: /obo/gecko/ihcc-gecko.owl
  replacement: ...
```

beckyjackson:
It looks like the `test.py` script only tests redirects. I think adding a test for names into this would be a bit weird (I'm not sure how to format the "failures"), but perhaps we should add another test script to check that the products are all `[IDSPACE].*`?

PR #692, opened 2020-10-01
Create gsso.yml

matentzn:
@Superraptor

### DIFF
--- a/config/gsso.yml
+++ b/config/gsso.yml
@@ -14,9 +14,6 @@ example_terms:
 
 entries:
 
-- exact: /
-  replacement: https://github.com/Superraptor/GSSO/
-
 - prefix: /releases/
   replacement: https://github.com/Superraptor/GSSO/releases/download/v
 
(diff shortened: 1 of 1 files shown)


config/gsso.yml
```diff
@@ -0,0 +1,33 @@
...
+- prefix: /about/
+  replacement: http://www.ontobee.org/ontology/GSSO?iri=http://purl.obolibrary.org/obo/
+
+- prefix: /subsets/
+  replacement: https://raw.githubusercontent.com/Superraptor/GSSO/master/subsets/
+
+- prefix: /
+  replacement: https://raw.githubusercontent.com/Superraptor/GSSO/master/
+
+- exact: /
```
jamesaoverton, line 32: Apache matched entries in order. This `exact: /` entry will never be matched, because the previous `prefix: /` hides it.

jamesaoverton (changes requested):

jamesaoverton (approved):

Superraptor (commented):
Everything appears to be in order! Thank you so much!

PR #693, opened 2020-10-07
Update MRO config

beckyjackson:
Due to an increase in size above 50GB, we have moved `mro.owl` out of the repository and now include it as a release binary. This PR updates the redirects for the location.

For dated releases:
We could update the `/20` prefix to point to the releases, but then we would lose everything before 2020-10-07. Perhaps we do that and then add those in as exact entries? Or we could add those products to the past releases. @jamesaoverton


jamesaoverton:
This is fine for now.

PR #702, opened 2020-11-03
Redirect DUO releases/ to list of releases

jamesaoverton:


matentzn (commented):
I know this is what Melanie said, but isn't what she really wants:

```
https://github.com/EBISPOT/DUO/releases/download/v
```

?

mcourtot:
I think @jamesaoverton has it right - sorry if I was unclear. I'd like to send to https://github.com/EBISPOT/DUO/releases so that users can see the list of all dated releases.

matentzn:
Note though that if you do that, your version iris will not resolve!

matentzn:
This main purpose of this redirect is so that you can use http://purl.obolibrary.org/obo/duo/releases/2020-02-03/duo.owl to obtain the right DUO version (OBO versioning principle).

matentzn:
(in any case I can see now the `exact`... so my comments pertain to the other, the `prefix` entry.

jamesaoverton:
@matentzn These are working for me http://purl.obolibrary.org/obo/duo/releases/2017-10-16/duo.owl redirects to a valid OWL file https://raw.githubusercontent.com/EBISPOT/DUO/v2017-10-16/duo.owl

The Travis check failed because my tests aren't quite clever enough. I'll fix that and merge.

matentzn:
Great! Didnt know that that works as well!

PR #708, opened 2020-11-24
Add UO purls for additional product/variants

reality:
addresses https://github.com/bio-ontology-research-group/unit-ontology/issues/23

### DIFF
--- a/config/uo.yml
+++ b/config/uo.yml
@@ -6,6 +6,12 @@ base_url: /obo/uo
 products:
 - uo.owl: https://raw.githubusercontent.com/bio-ontology-research-group/unit-ontology/master/uo.owl
 - uo.obo: https://raw.githubusercontent.com/bio-ontology-research-group/unit-ontology/master/unit.obo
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Although the validation currently accepts other values, the `products` should really be limited to `IDSPACE.FORMAT` to avoid conflicts with other ontology IDSPACEs. Can these new PURLs be put somewhere else in the UO namespace?

reality:
Thanks for the input. I would be happy with the links being anywhere as long as they are available from UO's PURL, however I'm not sure how to specify that in the file. Could you perhaps give an example of how to do that? Thank you!

jamesaoverton:
The [obi.yml](https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/obi.yml#L18) shows a range of examples. The first few `entries` are alternative products. For example:

```
- exact: /obi-base.owl
  replacement: https://raw.githubusercontent.com/obi-ontology/obi/master/views/obi-base.owl
```

defines this PURL <http://purl.obolibrary.org/obo/obi/obi-base.owl> which points to the latest [base](https://github.com/INCATools/ontology-development-kit/blob/master/docs/ReleaseArtefacts.md#release-artefact-1-base-required) file for OBI.

PR #717, opened 2020-12-29
Update to version 1.2

FrancescoVit:


matentzn:
Hey @FrancescoVit! We generally discourage the removal of purls in the PURL settings, because one of the main point, if not the main point, of PURLs is that once people refer to your ontology through a purl, they expect it to be there forever; However, it seems you never had an OBO product for ONS? In that case I think it would be fine to remove the entry as a kind of oversight in the first place.

FrancescoVit:
Hi @matentzn, totally understand the point. The ONS had a .obo for the previous version, but I currently removed it has I am having problems converting for the v1.2. Hence I could totally re-upload the .obo file of the v1.1 and adding the following 

- exact: /ons_release_v1.1.obo
  replacement: https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/ons_release_v1.1.obo

would that be ok?

jamesaoverton:
@FrancescoVit You could continue to link to the 1.1 version in `.obo` format.

My preference would be to fix the root problem. Do you have a GitHub issue for the conversion problem you're having, so we can help figure that out?

matentzn:
Yeah, I think ONS owl violates the OBO profile (multiple labels/comments etc):

```
OBO STRUCTURE ERROR Ontology does not conform to OBO structure rules:
multiple name tags not allowed. in frame:Frame(ONS:0000101 id( ONS:0000101)comment( Should break the definition down into separate terms, e.g. bioaccessable should be defined separately.)property_value( IAO:0000117 ENPADASI_WP4 xsd:string)name( food digestion)name( Food digestion)def( Digestion is the breakdown of large food molecules into small water-soluble food molecules. The breakdown is both mechanical and chemical (enzymatic digestion). During digestion, food components are released from the food matrix and may become bioaccessible.)def( Digestion is the breakdown of large insoluble food molecules into small water-soluble food molecules. The breakdown is both mechanical and chemical (enzymatic digestion). During digestion, food components are released from the food matrix and become bioaccessible. Bioaccessible molecules can be absorbed.)is_a( GO:0008150))
```

@FrancescoVit you can use

```
 robot convert -I https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/ons_release_v1.2.owl -f obo --check false -o ons_release_v1.2.obo
```

to circumvent the issue (--check false). This produces slightly dodgy OBO, but we all do it, all the time.

jamesaoverton:
Thanks @matentzn. I'd prefer to do this on the ONS tracker.

matentzn:
This particular error, by the way, should be fixed: you should never use multiple labels on a class. Use 

```
 robot report -I https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/ons_release_v1.2.owl -o report.txt
```

To identify such issues!

matentzn:
Made a ticket: https://github.com/enpadasi/Ontology-for-Nutritional-Studies/issues/9

jamesaoverton:
@FrancescoVit Would you like us to update the `ons.owl` to 1.2 now and update the `ons.obo` later? Or would you like to wait and update both at the same time?

FrancescoVit:
@jamesaoverton maybe the best would be to close this, solve all issues in ONS, and then modifying the yml at once? I am still fixing core problems in ONS for the obo format

jamesaoverton:
Ok, let us know if you need anything.

PR #718, opened 2021-01-06
Update pdumdv.yml

matentzn:
See https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1389


jamesaoverton:
This is breaking the test for the `example_terms`. I'm not sure how serious that is, since none of the project maintainers have every touched this file -- I made up that example myself. I don't know what to do.

matentzn:
I added an actual term from the ontology as an example.

PR #723, opened 2021-01-21
Get rid of the whole "current" tag logic in Mondo

matentzn:
Thanks to @jamesaoverton I now learned that you can actually reference the latest release without a current tag. This simplifies our release pipelines a lot!

@cmungall


matentzn:
I would like @cmungall to look this over in case the current tag served any other purpose.

cmungall (approved):

cmungall:
yes this does simplify things. do it!

PR #721, opened 2021-01-10
Update ontoneo.yml

FernandaFarinelli:
This is an update related to request  "New version of Ontoneo #720".

PR #739, opened 2021-02-25
Revised update for ontoneo

jamesaoverton:
See #721

I started with [this file](https://github.com/ontoneo-project/Ontoneo/blob/master/ontoneo.yml) and made some fixes.

### DIFF
--- a/config/ontoneo.yml
+++ b/config/ontoneo.yml
@@ -21,9 +21,9 @@ entries:
 
 ### Imported ontologies current release
 - prefix: /import/
-  replacement: https://raw.githubusercontent.com/ontoneo-project/Ontoneo/master/import/
+  replacement: https://raw.githubusercontent.com/ontoneo-project/Ontoneo/master/releases/2021-02-24/import/bfo_import.owl
   tests:
-  - from: /import/uberon_import.owl
+  - from: releases/2021-02-24/import/bfo_import.owl
     to: https://raw.githubusercontent.com/ontoneo-project/Ontoneo/master/import/uberon_import.owl
 
 ### Previous release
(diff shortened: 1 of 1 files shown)


jamesaoverton:
@FernandaFarinelli This version addresses my concerns from #721. Thanks. Can you please double-check this?

I also want to say that I'm a bit worried about how you're using the `prefix: /import/` rule. Projects do use these sorts of imports for their development versions, but it's much safer for your release file to be a single merged OWL file with no `owl:imports`.

FernandaFarinelli:
@jamesaoverton Thank you for your feedback. Sore for my long text.

About your worried about my imports, I import many parts of different OBO Foundry ontologies. Based on MIREOT, I thought that create owl files with just the set of reusing elements for each ontology, and then, import just these files.

I thought that separated files could be easy for someone that wants to use ONTONEO to understand. Such separation can decouple the dependency between the reused ontologies.

In my view, separated files could provide easy maintenance on import files, for instance, if a new element of OBI is needed, I just need to add this new element on the obi owl file. In some cases, the main file (ontoneo.owl) could be the same. At the same time, separated files could improve maintenance on ontoneo.owl file, in the way that this file is smaller.

For me, keeping the files separate can facilitate and clarify some ontology governance processes

What do you think about this? Does my view make sense? I appreciate your consulting in this case.

jamesaoverton:
I'm not asking for any changes to Ontoneo. Does this PURL config work for you?

I'm just making a suggestion that you can consider implementing sometime. It's based on OBO best practises, and it may make life easier for you and your users.

What you are saying makes sense. Yes, keep your imports separate for the *development* version of your ontology, for all the good reasons you laid out. People can see all those separate parts if they want to look on GitHub. But then I'm suggesting one more step beyond what you're saying: merge them all together into the *release* version of your ontology.

Many OBO projects use this pattern, but here's an example from OBI:

- we edit the [`src/ontology/obi-edit.owl`](https://github.com/obi-ontology/obi/blob/master/src/ontology/obi-edit.owl) file
- it has `owl:import`s pointing to various imports from other ontologies here: https://github.com/obi-ontology/obi/tree/master/src/ontology/OntoFox_outputs
- but the release version [`obi.owl`](https://github.com/obi-ontology/obi/blob/master/obi.owl) has been merged (and reasoned), so it contains all the terms (OBI terms and imported terms) without any `owl:import` statements

This will be much easier for tools such as term browsers and RDF triplestores to use, because they only have to load one file. It can be used offline, doesn't require pulling large files across the internet every time you open the file, and is generally safer from network problems. But most importantly, when a user loads an older version of OBI, they will see exactly what was in the ontology at the time of release, and not a mixture of old OBI terms with newer imports. That last point is my biggest concern about your current approach with Ontoneo.

I'm not asking for any changes to Ontoneo. As far as I'm concerned, this PR is ready to merge.

config/ontoneo.yml
```diff
@@ -21,24 +21,30 @@ entries:
 
 ### Imported ontologies current release
 - prefix: /import/
-  replacement: https://raw.githubusercontent.com/ontoneo-project/Ontoneo/master/import/
+  replacement: https://raw.githubusercontent.com/ontoneo-project/Ontoneo/master/releases/2021-02-24/import/
   tests:
-  - from: /import/uberon_import.owl
-    to: https://raw.githubusercontent.com/ontoneo-project/Ontoneo/master/import/uberon_import.owl
+  - from: /import/bfo_import.owl
+    to: https://raw.githubusercontent.com/ontoneo-project/Ontoneo/master/releases/2021-02-24/import/bfo_import.owl
```
matentzn, line 27: Doesnt this mean you have to update this with every release?

FernandaFarinelli: I fixed this in my new version.

PR #744, opened 2021-03-12
Update exo.yml

matentzn:
see https://github.com/CTDbase/exposure-ontology/issues/16

FYI @cjgrondin

### DIFF
--- a/config/exo.yml
+++ b/config/exo.yml
@@ -4,8 +4,8 @@ idspace: ExO
 base_url: /obo/exo
 
 products:
-- exo.owl: http://ontologies.berkeleybop.org/exo.owl
-- exo.obo: https://github.com/CTDbase/exposure-ontology/raw/master/src/ontology/exo.obo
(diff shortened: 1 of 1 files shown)


config/exo.yml
```diff
@@ -4,8 +4,8 @@ idspace: ExO
 base_url: /obo/exo
 
 products:
-- exo.owl: http://ontologies.berkeleybop.org/exo.owl
-- exo.obo: https://github.com/CTDbase/exposure-ontology/raw/master/src/ontology/exo.obo
+- exo.owl: https://github.com/CTDbase/exposure-ontology/raw/master/exo.obo
+- exo.obo: https://github.com/CTDbase/exposure-ontology/raw/master/exo.owl
```
jamesaoverton, line 8: These file extensions got swapped.

jamesaoverton (changes requested):

config/exo.yml
matentzn: What would I do without you!

jamesaoverton (approved):

PR #757, opened 2021-05-03
Update pco.yml

ramonawalls:
PCO release

### DIFF
--- a/config/pco.yml
+++ b/config/pco.yml
@@ -12,6 +12,8 @@ example_terms:
 - PCO_0000000
 
 entries:
+- exact: /releases/2021-05-03/pco.owl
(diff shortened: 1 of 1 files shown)


ramonawalls:
@jamesaoverton Did this fail due to issue #751? The only change I made was to add a new exact replacement for the latest version, which seems like something even I could get right.

jamesaoverton:
Yes, the failure is not your fault. Your change looks fine. I'll test it locally and merge it as soon as I get a chance.

ramonawalls:
Thanks @jamesaoverton .

cmungall:
@ramonawalls you could do this the same way we do for many other ontologies, where there is no need to update the cfg with each release

See ODK and for example:
https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/pato.yml#L21-L22

PR #762, opened 2021-05-18
Update ons.yml

FrancescoVit:


config/ons.yml
```diff
@@ -16,7 +17,10 @@ entries:
...
 - exact: /ons_release_v1.1.owl
-  replacement: https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/ons_release_v1.1.owl
+  replacement: https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/Older_deprecated/ons_release_v1.1.owl
+
+- exact: /ons_release_v1.0.obo
+  replacement: https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/Older_deprecated/ons_release_v1.1.obo
   
 - prefix: /imports/
-  replacement: https://github.com/enpadasi/Ontology-for-Nutritional-Studies/tree/master/src/ontology/imports
+  replacement: https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/imports/ONS_imports.owl
```
matentzn, line 26: This is probably not working. I dont think you need to make this change at all, or at least no include `ONS_imports.owl` here. In your ontology (ONS) you actually hard code the raw github location in the import (which you probably should not), but I would remove `ONS_imports.owl` and the purl to http://purl.obolibrary.org/obo/ons/imports/OMS_imports.owl will work just fine.

matentzn (changes requested):

config/ons.yml
FrancescoVit: Ok, I'll revert back to how it was

- prefix: /imports/
  replacement: https://github.com/enpadasi/Ontology-for-Nutritional-Studies/tree/master/src/ontology/imports

or you would suggest removing it entirely?

matentzn: I updated it now the way I think it should be! What do you think?

FrancescoVit (commented):
Ok the change is fine, thank you for the help. Getting the hang of a lot of things here, github wonders included.

FrancescoVit (commented):
ok

config/ons.yml
FrancescoVit: Ok the change is fine, thank you for the help. Getting the hang of a lot of things here, github wonders included.

config/ons.yml
```diff
@@ -23,4 +23,4 @@ entries:
   replacement: https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/Older_deprecated/ons_release_v1.1.obo
   
 - prefix: /imports/
-  replacement: https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/imports/ONS_imports.owl
+  replacement: https://raw.githubusercontent.com/enpadasi/Ontology-for-Nutritional-Studies/master/src/ontology/imports/
```
FrancescoVit, line 26: This should not point to "https://raw.githubusercontent.com" as now this is a folder

matentzn: Thats ok: this is not an exact purl to some file! It is a prefix pattern, so you wouldn't just try to go to /imports/, you would always look for a specific file like:

http://purl.obolibrary.org/obo/ons/imports/ONS_imports.owl

The purl rule here would make sure that if you look for that (the above example), it would redirect to raw github.

FrancescoVit: Ok thanks!

FrancescoVit (commented):
All seems ok now

matentzn (approved):

PR #764, opened 2021-05-26
Use owl version from VariO homepage

nicholsn:
The OWL version of VariO on [bioportal](https://bioportal.bioontology.org/ontologies/VARIO) and their [homepage](http://variationontology.org/download.shtml) is newer with additional content compared to the one at http://ontologies.berkeleybop.org/vario.owl

This PR sets the config to use the ontology distributed from the homepage.

Note that it looks like the ontology was updated, but the timestamp and version was not ([diff](https://data.bioontology.org/ontologies/VARIO/submissions/16/download_diff?apikey=8b5b7825-538d-40e0-9e9e-5ab9274a9aeb). It also appears that the Version IRI (http://purl.obolibrary.org/obo/vario/2018-11-09/vario.owl) is missing from the homepage distribution but present in the berkeleybop distribution.

cc @MattBrauer

### DIFF
--- a/config/vario.yml
+++ b/config/vario.yml
@@ -4,7 +4,7 @@ idspace: VariO
 base_url: /obo/vario
 
 products:
-- vario.owl: http://ontologies.berkeleybop.org/vario.owl
+- vario.owl: http://www.variationontology.org/vario_download/vario.owl
 - vario.obo: http://www.variationontology.org/vario_download/vario.obo
 
 term_browser: ontobee


jamesaoverton:
Thanks @nicholsn. This PR is fine, and I'll merge it.

I have two comments, though:

1. VariO is marked as "orphaned" in the OBO registry. Are you responsible for the ontology? If not, who is? Should it be marked as an "active" ontology in the registry?

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/ontology/vario.md

This issue says that the OBO registry team is waiting to hear what license VariO is shared under:

https://github.com/OBOFoundry/OBOFoundry.github.io/issues/886

2. [OBO Principle 2: Common Format](http://obofoundry.org/principles/fp-002-format.html) requires that the main ontology product (`vario.owl` in this case) be published in RDFXML format, which is the most widely supported ontology format. However the `vario.owl` file is actually in OWL Functional Syntax format. None of the downloads [here](http://www.variationontology.org/download.shtml) are in RDFXML format. Would it be possible for VariO to publish RDFXML format., or would that break stuff?

jamesaoverton (approved):

MattBrauer:
There is a 2020 paper that refers to this ontology: https://www.tandfonline.com/doi/full/10.1080/15476286.2020.1817266

I've reached out to the author about this issue and will follow up to gauge his intent for maintenance of the ontology.

nicholsn:
Hey @jamesaoverton, thanks for the merge!  I'm not the maintainer, the contact listed is Mauno Vihinen that @MattBrauer reached out to. 

The motivation for this is to get dereferencing to work for terms in the newer version. For example, http://purl.obolibrary.org/obo/VariO_0504 is present in [Bioportal](https://bioportal.bioontology.org/ontologies/VARIO?p=classes&conceptid=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FVariO_0504) but does not show up in Ontobee or OLS.

Do you know if Ontobee will update based on the change from this PR?

jamesaoverton:
Thanks @MattBrauer!

@nicholsn Yes, Ontobee should update automatically, but I'm not sure what their update schedule is. I believe that it's weekly.

Issue #754, opened 2021-04-28
Create a shared docker-based infrastructure for the deploying the OBO PURL server


kltm:
Currently, the OBO PURL server is held on a privately held AWS EC2 instance. We'd like to move this into commonly held infrastructure (AWS EC2, IAM, on GO master payer) based on docker images.

Tackling this issue would include:

- [x] a docker image for the PURL server that could be used for development and production
- [x] a well-tested automated system (ansible?) or v/tight SOP doc for deploying a production PURL server with correct DNS
- [x] includes adding in the fixes from #63 
- [x] Setting the appropriate IAM users and groups to give all parties sufficient access to accomplish the tasks
- [x] distribute ssh keys

I might also suggest:

- [ ] external uptime monitoring. Uptimerobot? Something else? CloudWatch?
- [ ] a failback server (@jamesaoverton 's old server?), in case we're nervous

Stages of development may include:

- [x] getting a local docker instance of the PURL server working
- [x] updating current travis to GHA using this image
- [x] getting a docker-based instance working on AWS EC2
- [x] getting a docker-based instance working on AWS EC2 under "dummy domains"
- [x] aim production DNS at new system
- [x] confirm automatic deployment elements are in place

Tagging @jamesaoverton @cmungall for feedback on the above before moving ahead.

kltm:
Noting work for logs/analysis fat #63 on https://github.com/OBOFoundry/purl.obolibrary.org/pull/747

kltm:
@jamesaoverton We're thinking about starting a Dockerfile from scratch, but I seem to recall that you may have already started on something? Is there anything worth basing work from that already exists?

jamesaoverton:
@kltm I just pushed this [docker-experiments](https://github.com/OBOFoundry/purl.obolibrary.org/tree/docker-experiments) branch.

kltm:
@jamesaoverton Great, thank you!
Talking to @cmungall , it sounds like this ticket is the common understanding of our recent conversation, so I'm going to move it into our "ready" hopper.
Tagging @abessiari for further discussion as we progress.

jamesaoverton:
A bit more...

My top priority is to replace Travis, which is broken for this repo, with GitHub Actions, so we have CI for PRs again. The experiment branch was working toward that, but then I got clobbered by other deadlines.

About my `docker-experiment` branch:

- `tools/site.yml`: the Ansible script we've been using, which should still work
[...]

PR #765, opened 2021-05-26
issue 83 logs and analysis using docker

abessiari:
- Docker file using ubuntu 20.04 and python 3.8.5
- Support for s3 buckets. 
- See README under docker directory. 
- Minimal modification to ansible scripts to support ansible 2.9.6
- Tested using:  sudo make all test; sudo make safe-update
- Tested Log rotation using: sudo :logrotate -v -f /etc/logrotate.d/apache2
- Added  github workflow build_and_test.yml


abessiari:
@jamesaoverton 
cc: @kltm 

Hi James this pull request supersedes PR 747.
Docker support has been added. 

The only thing that I could not test is make safe-update as it keeps saying that the last build failed and thus no updates.

It would be nice if you can take a look and help test/make sure we are on the right path.
Thanks.
aes

jamesaoverton:
Thanks! Yes, safe-update is completely broken because we ran out of free Travis credits.

I will look at this and test is as soon as I can, which unfortunately won't be for a few days.

kltm:
Linking https://github.com/OBOFoundry/purl.obolibrary.org/issues/754

aws/vars.tf
```diff
@@ -0,0 +1,32 @@
+variable "tags" {
+  type = map
+  default = { Name = "test-purl-obolibrary" }
+}
+
+variable "instance_type" {
+  default = "t2.large"
```
jamesaoverton, line 7: We've been getting by with `t2.micro` with no problems.

aws/vars.tf
```diff
@@ -0,0 +1,32 @@
...
+variable "ssh_port" {
+  type        = number
+  default     = 22
+  description = "ssh server port"
+}
+
+variable "http_port" {
+  type        = number
+  default     = 80
+  description = "noctua server port"
```
jamesaoverton, line 31: There shouldn't be any references to Noctua.

aws/ssh_key.tf
```diff
@@ -0,0 +1,5 @@
+resource "aws_key_pair" "ssh_key" {
+  key_name   = var.key_name
+  public_key = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC1TL5/v2/fy0kyX68/W8YT2bML7XvK6IF84vgSYDCqk2E/UbsLXYbUEk5/nxFOBW+4gEmQP9D80jkAAOEN4zXalctKokib8SMHhCSzv+epCtqsbxKzU71Gfl9elJWNWZcwnCzwkZrO6RDZeu7Jcp3Asx4NiyAKd7PjWuLU8qmJo5rjwbAS4dumvlDLdjVWka4jRT02jVP5+p2J/+JwvZJCGMlEb6d4WKuy8UsO4wZXXzgwrYRPfBu3XaZFA19MoLAO1UyzZwKSsin7yg/+JN8IE7sIwvvL9pExD4DU0krfNyjAO8c88UhzrUbN/vEGxDXEIvvmIQ8k1MtHh8o6WwP3 AEssiari@aessiari-mba"
```
jamesaoverton, line 3: Who should have keys here?

dependencies/bare-metal.yml
```diff
@@ -0,0 +1,12 @@
+name: test_purl_obolibrary
+channels:
+  - defaults
+  - conda-forge
+  - anaconda
```
jamesaoverton, line 5: Why do we need Anaconda? Why use Python 2? The PURL system requires Python 3: https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/Makefile#L16

jamesaoverton:
Sorry for the delay in replying. There's a major grant deadline this week that has taken up all of my time.

I was able to build the Docker image and run `make` inside it. I didn't try the logrotate stuff.

I guess I have a bunch of questions about the deployment story. I don't understand why all the `aws/` files are here.

dependencies/bare-metal.yml
jamesaoverton: I guess this is for local testing on macOS?

jamesaoverton:
It's really important to have CI on this repo. We ran out of Travis credits. The plan has been to use GitHub Actions but I though we'd wait for this. What is the plan for using this work with GitHub Actions?

abessiari:
@jamesaoverton 

Thanks for looking into this.  
The aws stuff was a copy and paste from noctua. I will clean it up. 
I will take a look at the conda file as well.

abessiari:
@jamesaoverton 
I will just remove the aws files and conda. Initially I was testing on aws. So I just thought to squezze them in. :-)

abessiari:
@jamesaoverton 

I have removed the aws and the conda files. They are not needed at this point. 
For Github Actions please advise (some details) on what needs to happen if you want me to take a crack at it.

jamesaoverton:
Thanks for those changes @abessiari. I just haven't been able to find a couple of uninterrupted hours to work on this. I'll keep trying.

abessiari:
@jamesaoverton 
@kltm 

Hi. 

I added a build_and_test.yml github workflow.  
This builds the docker image and runs the test. 
It is under .github/workflows. 

Next steps is to use this for safe_update. Can you and I have a zoom meeting 
to discuss how to go about this. 
Thanks.
aes

jamesaoverton:
Thanks @abessiari, much appreciated. I'm happy to talk with you on Zoom. Let's arrange that on the OBO community Slack, if you're on it, or just email me: james at overton.ca.

The current approach to "safe-update" is a "pull" approach for updating the PURLs. We use `cron` every 10 minutes to check for an update that has passed the CI tests. We could use a "push" approach instead, with a webhook or something. Or if we're packaging and deploying via Docker then I guess we want another approach entirely? @kltm?


If we do want to update the current "safe-update" approach, we need to make some changes. Basically tools/safe-update.py script queries the Travis CI API to see whether the master branch CI tests have succeeded. If they did succeed, we update the PURLs. If not, we don't. We aren't using Travis any more, but we still want to check the status of the GitHub Actions for the master branch.

This should be simple. For a given check like <https://github.com/obi-ontology/obi/actions/workflows/obi-test.yml> there's an easy way to get an SVG status badge. But I can't figure out a similarly easy way to get a pass/fail exit status.

This is ugly but it returns "success" for the OBI repository's master branch:

```sh
gh api repos/obi-ontology/obi/commits/$(gh api repos/obi-ontology/obi/branches/master | jq -r .commit.sha)/check-runs | jq .check_runs[0].conclusion
```

The [GitHub CLI](https://cli.github.com/) should be able to just do this, but I can't figure it out. Does anybody have a better approach?

kltm:
Overall, I'd suggest that we treat the these as two separate problems: 1) getting everything dockerized and turnkey so that anybody with enough permissions in AWS (EC2 and Route 53) can bring up a new functioning instance and 2) updating a running system when new correct metadata exists.
I think that we'd probably end up spending a lot of time and energy making a seamless system if 1 and 2 were mixed--so that's a vote from me maybe for exploring some kind of "safe-update" first. For solving just 2, I think a publicly held and modifiable GHA file would be great, but having a clever cron job like you had before or using the API like you have above would be fine as well (or a mix).

abessiari:
@jamesaoverton 
cc @kltm 

So I have checked in tools/safa-update.py that uses github api instead of travis. We can chat tomorrow about it in our meeting at 9:00 am.

abessiari:
@jamesaoverton 

Ok we are ready to merge. 
Once the merge happens I will continue testing using the  master repo. 
Thanks.

jamesaoverton:
Sorry for the delay. Thanks for all your work on this!

jamesaoverton:
Thanks for all your work on this!

abessiari:
Thanks.

PR #768, opened 2021-07-21
Update uberon.yml

matentzn:


jamesaoverton:
Unfortunately PURL server updates are manual at the moment, but I pulled this and it should be live now.

matentzn:
Thanks @jamesaoverton! Sorry about that!

PR #769, opened 2021-07-27
Add history of gno releases

Hendricks27:


config/gno.yml
```diff
@@ -33,7 +33,21 @@ entries:
...
+  - exact: /2020-10-14/gno.owl
+    replacement: https://github.com/glygen-glycan-data/GNOme/releases/download/V1.6.3/GNOme.owl
+
+  - exact: /2021-03-05/gno.owl
+    replacement: https://github.com/glygen-glycan-data/GNOme/releases/download/V1.7.0/GNOme.owl
+
+  - exact: /2021-05-12/gno.owl
+    replacement: https://github.com/glygen-glycan-data/GNOme/releases/download/V1.7.1/GNOme.owl
+
+  - exact: /2021-6-16/gno.owl
```
jamesaoverton, line 52: In this entry the month "6" is just one digit, while the others are two digits. Is that intentional?

Sorry to nitpick but small things like this can cause problems later. Otherwise this PR looks good to me.

Hendricks27: Thanks for catching that! Sorry for the bug.
It is fixed now.

PR #774, opened 2021-08-05
Switch order of bds prefix entries

jamesaoverton:
FYI @hkir-dev: PURL entries match in order, so the `/bds/` entry was matching before `/bds/browser/` entry, and `/bds/browser/` could never be reached. I swapped the order of the two entries to fix this.

I missed this when I reviewed the PR, and the automated tests didn't catch it either, which is not how it's supposed to work.


hkir-dev:
Thank you for the feedback and the bug-fix.

PR #789, opened 2021-09-22
Update bspo.yml

matentzn:


config/bspo.yml
```diff
@@ -6,6 +6,7 @@ base_url: /obo/bspo
 products:
 - bspo.owl: https://raw.githubusercontent.com/obophenotype/biological-spatial-ontology/master/bspo.owl
 - bspo.obo: https://raw.githubusercontent.com/obophenotype/biological-spatial-ontology/master/bspo.obo
+- bspo.json: https://raw.githubusercontent.com/obophenotype/biological-spatial-ontology/master/bspo.json
```
jamesaoverton, line 9: This looks correct, but I get a 404. Should I go ahead anyway?

matentzn: Oh sorry, fixed in https://github.com/obophenotype/biological-spatial-ontology/pull/31

matentzn: Fixed now, thank you!

PR #796, opened 2021-11-01
experiment: redirect to amigo

cmungall:
I suspect this will not work - amigo only works with CURIEs

### DIFF
--- a/config/go.yml
+++ b/config/go.yml
@@ -10,9 +10,10 @@ products:
 
 term_browser: custom
 example_terms:
-- GO_0000001
+- GO_0006915
 
 entries:
+
 - exact: /go.obo
   replacement: http://current.geneontology.org/ontology/go.obo
 
--- a/config/go.yml
+++ b/config/go.yml
@@ -8,7 +8,7 @@ products:
 - go.obo: http://current.geneontology.org/ontology/go.obo
 - go.json: http://current.geneontology.org/ontology/go.json
 
-term_browser: ontobee
+term_browser: custom
 example_terms:
 - GO_0000001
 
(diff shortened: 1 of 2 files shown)


jamesaoverton:
You need to define the custom redirect in [`obo.yml`](https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/obo.yml#L49)

cmungall:
Thanks! Will do.

cmungall:
Aside: there is some confusion redundancy in some of the ymls; e.g. hp.yml has

```yaml
- prefix: /about/
  replacement: http://www.ontobee.org/browser/rdf.php?o=HP&iri=http://purl.obolibrary.org/obo/
  tests:
  - from: /about/HP_0000001
    to: http://www.ontobee.org/browser/rdf.php?o=HP&iri=http://purl.obolibrary.org/obo/HP_0000001
```

This is harmless but confusing, as it is over-ridden by the following in obo.yml

```
# Term redirects for HP
- regex: ^/obo/HP_(\d+)$
  replacement: https://hpo.jax.org/app/browse/term/HP:$1
  status: see other
  tests:
  - from: /HP_0002088
    to: https://hpo.jax.org/app/browse/term/HP:0002088
```

not a priority and likely noted elsewhere

jamesaoverton:
Looking at the git blame, that chunk of hp.yml is from the original migration from OCLC XML dumps to our YAML.

jamesaoverton (approved):
This change looks right to me.

Currently someone (i.e. me) has to manually update the PURL server after merging.

Are you ready for me to make this update any time, or do you want to coordinate the change so you can monitor?

cmungall:
I am OK for you to make the change any time but if you could notify @kltm and me after you do it on slack that would be great

PR #798, opened 2021-11-08
Create epso.yaml

phwegner:


jamesaoverton:
Thanks. I fixed two things. We'll wait on https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1646

matentzn:
@phwegner This seems like a _very_ minimal config.. Did you take a look around for other options such as adding an issue tracker purl etc? See here for a small example you could get inspired from:

https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/aism.yml

phwegner:
@matentzn  I was reading the documentation about that config file and filled everything that was required. I will try to add more information.

phwegner:
@matentzn I just edited the yaml file and added issue tracker etc.

matentzn (approved):

PR #802, opened 2021-11-19
Rename epso.yaml to epso.yml

matentzn:
How did we not notice that :P


phwegner (approved):

phwegner:
@matentzn Thank you for noticing that I changed it.

PR #804, opened 2021-11-24
Update epso.yml

phwegner:


matentzn (approved):

jamesaoverton:
I'm confused. Why do we also have #805 ?

matentzn:
This PR is redundant with #805 and should therefore be closed.

PR #805, opened 2021-11-24
Update epso.yml

phwegner:


config/epso.yml
```diff
@@ -2,8 +2,8 @@ idspace: EPSO
 base_url: /obo/epso
 
 products:
-- epso.owl: https://raw.githubusercontent.com/phwegner/Epilepsyontology/main/EPSO.owl
-- EPSO_merged.owl: https://github.com/phwegner/Epilepsyontology/tree/main/release/EPSO_merged.owl
```
matentzn, line 6: You are deleting the EPSO_merged entry here. Is that intended?

phwegner: Yes since the "product" is the epso.owl which is the merged ontology.

matentzn: Because you _just_ registered I think it is fine, but you should know that the P in PURL stands for "persistent" - so removing an entry like EPSO_merged.owl is a violation of that - Imagine a user relying on http://purl.obolibrary.org/obo/EPSO_merged.owl - a PURL, will now have their pipelines fail because of this change.

But, because you are so new and its highly unlikely you already have dependencies on EPSO_merged.owl, I think its fine.

phwegner: @matentzn  Okay that makes sense but while testing I realized the PURL pointed to a version that was not working anyway since it was the non merged version so nobody should have a dependency on that. But now everything is working.
But that brings me to a question. How do we deal with updates of the ontology since then again we are changing whats behind the PURL ?

matentzn: Changing what is behind the PURL is totally fine! So basically 

http://purl.obolibrary.org/obo/epso.owl

means: This is the latest version of epso. When there is a new version, whatever is behind this changes.

That is why we need version iris! This way people can see what version they are looking at. So this PR is correct now?

phwegner: Okay then everything is clear. Yes now the PURL is correct.

PR #811, opened 2021-12-16
Update and rename epso.yml to epio.yml

phwegner:
Updated Repository. The way the purl works is not affected since it is an exact copy of the Ontology.


jamesaoverton:
Waiting on a decision here: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1703

config/epio.yml
```diff
@@ -0,0 +1,25 @@
...
+
+term_browser: ontobee
+
+example_terms:
+- EPIO_0000011
+
+entries:
+
+- prefix: /releases/
+  replacement: https://raw.githubusercontent.com/SCAI-BIO/EpilepsyOntology/main/release
```
matentzn, line 16: The point of the "releases" entry is to enable versioned releases. You are just redirecting to a release folder. If you want this to work properly, you should use formal Github releases ([example](https://github.com/pato-ontology/pato/releases)), or at the very least, create _dated subfolders_ in the release directory, like:

release/2021-12-12/EPIO_merged.owl

Note if you do that, you also have to add epio.owl specifically to release folder, else the version purls wont work (i.e. http://purl.obolibrary.org/obo/epio/releases/2021-12-12/epio.owl).

matentzn (changes requested):
Apart from dealing with the releases issue, please check the other changes I performed in https://github.com/OBOFoundry/purl.obolibrary.org/pull/811/commits/d407fcdb98710c2880fe63a6ee107fb958e9d0b4

config/epio.yml
phwegner: But we did not have a release folder in the "EPSO" version is this now mandatory to have release folders ?

matentzn: No, its not mandated - but the purpose of the /releases/ section in the purl config is to provide redirects for version IRIs as the one above. You can simply remove this section if you want (although I don't recommend it) - we try to encourage people to not only have version IRIs (this is actually _required_, you cant pass the dashboard without, but you are doing that), but also have the resolve (i.e. when putting the version IRI in your browser, you should find the ontology of that version). What do you think?

phwegner: Okay if it's not mandatory I would like to leave it how it is since we have to get this running as soon as possible due to upcoming papers about the EPIO. But I'll discuss this with my team in our next weekly.

matentzn: Ok, so @jamesaoverton when you see this, you can either: 

- merge this as is
- remove the releases section

I leave this up to you, as you may have a different perspective on the matter. I requested clarification from the editorial team here: https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1707

PR #814, opened 2022-01-05
Add special rule for RO terms which sometimes include alpha chars

cmungall:
Fixes https://github.com/oborel/obo-relations/issues/94


jamesaoverton:
If you are changing RO to use OLS, I presume that you also want to change the RO's BFO terms to use OLS? 8e858f1e8e3ca1bfa2dcae6cfe436309bf510449

cmungall:
Thanks, yes!

PR #817, opened 2022-01-20
creating elastic ip. using init=true in docker compose. added start a…

abessiari:
- added init=true to avoid defunct processes
- added start/down services  playbooks
- create elastic ip instead of pointing one 
- allowing to run github workflow  build_and_test.yml (workflow_dispatch)


abessiari:
@jamesaoverton 
@kltm 

Hi James.
Changes concern deploying to AWS. Checks have passed .
Please merge.
Thanks.

PR #823, opened 2022-02-09
Production documentation

abessiari:
- Documentation/bash script  for provisioning a new purl server and destroying the old server

see  PRODUCTION_PROVISION_README.md


abessiari:
@jamesaoverton 
@kltm 

Please find the documentation to re-provision  a production purl-server when needed. I do recommend to  try provisioning without using s3 backend and workspaces first by following provision/PROVISION_AWS_README.md

kltm:
For https://github.com/OBOFoundry/purl.obolibrary.org/issues/754
This looks good to me, I would think it's fine to merge as it doesn't really change anything active in the core, but want to leave the button pressing to @jamesaoverton

jamesaoverton:
I agree with @kltm. I haven’t tested this but I will merge.

PR #825, opened 2022-02-16
Add HPOA annotation files to releases

matentzn:


config/hp.yml
```diff
@@ -19,28 +19,24 @@ entries:
 - prefix: /tracker
   replacement: https://github.com/obophenotype/human-phenotype-ontology/issues
 
+- prefix: /releases/hpoa/
+  replacement: https://github.com/obophenotype/human-phenotype-ontology/releases/
```
matentzn, line 23: This says: if you see a PURL such as:

http://purl.obolibrary.org/obo/hp/releases/hpoa/2020-02-14/genes_to_phenotype.txt

redirect to the respective github release.

config/hp.yml
```diff
@@ -13,6 +13,12 @@ products:
 term_browser: custom
 
 entries:
+- prefix: /releases/hpoa/
+  replacement: https://github.com/obophenotype/human-phenotype-ontology/releases/download/v
+  tests:
+  - from: /releases/hpoa/2022-02-14/genes_to_phenotype.txt
+    to: https://github.com/obophenotype/human-phenotype-ontology/releases/download/v2022-02-14/genes_to_phenotype.txt
```
drseb, line 20: what does this do?

drseb (approved):
Looks good. Let's merge and test

config/hp.yml
matentzn: the test is just for OBO PURL system to have something to check whether the redirect works the way intended. Not affecting anything around Purls, users etc..

PR #842, opened 2022-05-06
Update ms.yml

mobiusklein:
Attempt to update versioned redirects

### DIFF
--- a/config/ms.yml
+++ b/config/ms.yml
@@ -13,8 +13,8 @@ example_terms:
 - MS_0000000
 
 entries:
-- prefix: /releases/
-  replacement: https://raw.githubusercontent.com/HUPO-PSI/psi-ms-CV/v
+- regex: /releases/v?(\d+\.\d+\.\d+)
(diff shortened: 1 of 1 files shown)


jamesaoverton:
I'm worried about this change. It will not work for release PURLs for `.owl` files such as http://purl.obolibrary.org/obo/ms/releases/4.1.86/psi-ms.owl

I think you want:

```
- prefix: /releases/
  replacement: https://github.com/HUPO-PSI/psi-ms-CV/releases/download/v
  tests:  
  - from: /releases/4.1.86/psi-ms.obo
    to: https://github.com/HUPO-PSI/psi-ms-CV/releases/download/v4.1.86/psi-ms.obo
```

mobiusklein:
I can make those changes. I couldn't find an obvious example or template to follow and wanted to start with small changes.

matentzn:
What about this: https://github.com/OBOFoundry/purl.obolibrary.org/pull/845

PR #843, opened 2022-05-06
Fix links to examples in README.md

mobiusklein:
The links to the "test2" example in the README were broken, this repairs them to reference the correct location.


matentzn (approved):
Thank you very much @mobiusklein

PR #845, opened 2022-05-06
Update ms.yml

matentzn:
Get @mobiusklein review before merging.

### DIFF
--- a/config/ms.yml
+++ b/config/ms.yml
@@ -4,8 +4,8 @@ idspace: MS
 base_url: /obo/ms
 
 products:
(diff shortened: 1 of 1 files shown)


mobiusklein:
Looks good to me, assuming I follow this correctly,

`http://purl.obolibrary.org/obo/ms/` is the root address, and then `prefix` rules replace everything up to and including the prefix with a the `replacement` content verbatim (no regex captures needed). Then this remaps both OBO and OWL requests correctly.

The behavior seems right to me.

PR #847, opened 2022-05-13
Update omo.yml

zhengj2007:
Added removed version IRI back in

### DIFF
--- a/config/omo.yml
+++ b/config/omo.yml
@@ -11,6 +11,13 @@ term_browser: ontobee
 entries:
 - exact: /omo.owl
   replacement: https://raw.githubusercontent.com/information-artifact-ontology/ontology-metadata/master/omo.owl
+  
+# version IRIs
+- exact: /2020-06-08/omo.owl
+  replacement: https://raw.githubusercontent.com/information-artifact-ontology/ontology-metadata/v2020-06-08/omo.owl
+
+- exact: /2020-05-08/omo.owl
+  replacement: https://raw.githubusercontent.com/information-artifact-ontology/ontology-metadata/v2020-05-08/omo.owl
 
 # development version
 - exact: /omo-edit.owl


matentzn:
Thank you for catching that @jamesaoverton.

Yes, @zhengj2007 the Version IRIs follow the `releases` pattern, see [here](https://github.com/information-artifact-ontology/ontology-metadata/blob/master/omo.owl#L15).

The version IRI resolves fine, and we do not have to add it to config manually ever again: http://purl.obolibrary.org/obo/omo/releases/2022-04-27/omo.owl

Note that the these past versions that @zhengj2007 added here followed a different pattern: https://github.com/information-artifact-ontology/ontology-metadata/commit/3951ba18423423498584e2dd24f67e5514dd6b6f

So there actually is a case to merge this pull request as a one-off to ensure these two past version IRIs resolve.

matentzn:
Thank you for catching that @jamesaoverton.

Yes, @zhengj2007 the Version IRIs follow the `releases` pattern, see [here](https://github.com/information-artifact-ontology/ontology-metadata/blob/master/omo.owl#L15).

The version IRI resolves fine, and we do not have to add it to config manually ever again: http://purl.obolibrary.org/obo/omo/releases/2022-04-27/omo.owl

Note that the these past versions that @zhengj2007 added here followed a different pattern: https://github.com/information-artifact-ontology/ontology-metadata/commit/3951ba18423423498584e2dd24f67e5514dd6b6f

So there actually is a case to merge this pull request as a one-off to ensure these two past version IRIs resolve.

matentzn:
@zhengj2007 as an aside (I also adhere to that in 99.9% of the cases) - _no one should merge changes to the purl config other than James_.. This avoids for example miscommunication leading to a premature merge which breaks the whole PURL system.

PR #848, opened 2022-05-13
Change DOID term browser to disease-ontology.org

allenbaron:
@lschriml asked that DOID term URLs be redirected to disease-ontology.org. _I believe_ this implements that change.

Is there a way to make OLS a backup, in case disease-ontology.org is down for some reason? Could this be implemented by setting DO's term_browser to 'ols' instead of 'custom'?

Thanks.


jamesaoverton:
I got confirmation from @lschriml via Slack, so I've merged.

Issue #855, opened 2022-06-10
Update RBO yaml config


DanBerrios:
Remove OBO format product which is currently erroring out in our ODK

jamesaoverton:
I think this was resolved by #856

PR #856, opened 2022-06-10
Update rbo.yml

DanBerrios:
See issue #855.  Removes OBO formatted RBO as product. Also updated the URLs for the repo to point to the GH org-hosted URLs instead of my personal GH space URLs (they are redirected already by GH but would rather they show the org).

### DIFF
--- a/config/rbo.yml
+++ b/config/rbo.yml
@@ -5,6 +5,7 @@ base_url: /obo/rbo
 
 products:
 - rbo.owl: https://raw.githubusercontent.com/Radiobiology-Informatics-Consortium/RBO/master/rbo.owl
+- rbo.obo: https://raw.githubusercontent.com/DanBerrios/RBO/master/rbo.obo
 
 term_browser: ontobee
 example_terms:
--- a/config/rbo.yml
+++ b/config/rbo.yml
@@ -4,8 +4,7 @@ idspace: RBO
 base_url: /obo/rbo
 
 products:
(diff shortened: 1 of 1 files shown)


jamesaoverton:
The point of PURLs is that they're permanent, so I do not want to just drop the `rbo.obo` PURL. I suggest keeping it pointed to the older version of the file until the issue is fixed. Is that OK with you?

DanBerrios:
The problem we are having is that Ontobee will not pick up the latest version of our OWL file from our GH repo and Nico suggested that maybe the yaml entry for the OBO format is the cause, I guess because the version tag in that file does not match the latest version.

jamesaoverton:
I don't think that having a PURL for `rbo.obo` can be causing your problem with Ontobee. I merged a version that maintains the old `rbo.obo` PURL, and I'll follow up with Oliver.

DanBerrios:
@jamesaoverton  Thank you, our database curators rely on the purls working every day, so being without them has caused a lot of inefficiency in their workflows...

jamesaoverton:
RBO is loaded into OLS: https://www.ebi.ac.uk/ols/ontologies/rbo

If you prefer, you can redirect <http://purl.obolibrary.org/obo/RBO_> term IRIs to OLS instead of Ontobee using `term_browser: ols`. Ontobee is the default, but projects can choose their preferred ontology browser.

matentzn:
There is no way the OBO purl has anything to do with the update process in ontobee! Something got misunderstood somewhere.

PR #857, opened 2022-06-20
Create ADO.yml

BideZ:
We have created ADO.yml. 
This ADO ontology refers to the one submitted in https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1806

### DIFF
--- a/config/ADO.yml
+++ b/config/ADO.yml
@@ -6,14 +6,14 @@ products:
 
 
 term_browser: ontobee
-
+example_terms:
+- ADO_0000001
 
 entries:
 
 - prefix: /releases/
-  replacement: https://github.com/Fraunhofer-SCAI-Applied-Semantics/ADO/tree/main/Releases
+  replacement: https://raw.githubusercontent.com/Fraunhofer-SCAI-Applied-Semantics/ADO/main/Releases/2022-06-11/
   
-
 - prefix: /tracker/
   replacement: https://github.com/Fraunhofer-SCAI-Applied-Semantics/ADO/issues
 
--- a/config/ADO.yml
+++ b/config/ADO.yml
@@ -1,0 +1,22 @@
+idspace: ADO
+base_url: /obo/ado
+
+products:
+- ado.owl: https://raw.githubusercontent.com/Fraunhofer-SCAI-Applied-Semantics/ADO/main/Releases/2.0.0/ADO.owl
+
+
+term_browser: ontobee
+
+
+entries:
+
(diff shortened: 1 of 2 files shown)


matentzn:
- [ ] Blocked by https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1964#issuecomment-1160372564

config/ADO.yml
```diff
@@ -0,0 +1,22 @@
+idspace: ADO
+base_url: /obo/ado
+
+products:
+- ado.owl: https://raw.githubusercontent.com/Fraunhofer-SCAI-Applied-Semantics/ADO/main/Releases/2.0.0/ADO.owl
```
matentzn, line 5: This is a redirect to a version iri. There should be one versioned release, and one non versioned release (you don't want to have to edit this file every time you create a new version.

I would suggest this:

1. You remove `Releases/2.0.0/ADO.owl` and simply add a `ado.owl` (lower case!)
2. then whenever you make a release, you run a github release like you do [here](https://github.com/Fraunhofer-SCAI-Applied-Semantics/ADO/releases)
3. You can then simply refer to ado.owl through the GitHub release tag in this purl config (I will show you how)

matentzn (changes requested):

config/ADO.yml
```diff
@@ -0,0 +1,22 @@
+idspace: ADO
```
matentzn, line 1: filename should be lower case (ado.yml)

matentzn: idspace should be upper case (dont change it!)

matentzn (changes requested):

config/ADO.yml
BideZ: Thank you for your reply. I have changed the file name as well as the products link. The link does not resolve now.

config/ADO.yml
matentzn: Great thx

config/ADO.yml
matentzn: Ok, one more thing required, and then you are done:

- [ ] move Releases/2.0.0/ADO.owl in your repo to ado.owl (top level) - delete the `Releases` folder altogether because GitHub releases will take care of your versioning.

matentzn: - [ ] ado.owl must be lower case.
- [ ] you cant just delete a release, you also have to delete the associated tag. then redo the release.

matentzn (approved):
Ready for review by @jamesaoverton

PR #864, opened 2022-08-30
add simple and base to purl redirect

shawntanzk:
I *think* I did this right? @matentzn - would you mind helping me check please? Thanks :)


shawntanzk:
Seems like the redirect is http://purl.obolibrary.org/obo/cl/cl-simple.owl - Not too sure if we still want to keep the redirect in this PR?

matentzn (approved):
yep, all ok!

PR #866, opened 2022-09-03
Add obophenotype slim to Chebi PURL config

matentzn:


jamesaoverton:
Can we close this?

matentzn:
I will put it on my todolist for December.

matentzn:
This raises a bit the question of "who owns the PURL config" - I personally think this is ok, especially for such an important subset that we need to release/manage ourselves rather than have CHEBI manage it. Formally, this may need a bit of squinting, but practically, I think this makes sense.

jamesaoverton:
As a first pass, I would think that the question "who owns the PURL config?" would have the same answer as "who owns the OBO registry entry?". As usual, ChEBI is probably an exception.

I can see the benefits here, but my gut feeling is that this change is unnecessary and sets a bad precedent. Isn't there another, better home for these PURLs?

matentzn:
I am fine putting it anywhere, but where? Outside of OBO you mean? These are literal subsets of obo ontologies, like RO core or similar..

cmungall:
I think it's misleading to have this in the chebi namespace if the releases are not 100% synced

matentzn:
Its in any case late now. Dozens of ontologies use the github raw link on the `main` branch so.. Lets close.

PR #868, opened 2022-09-15
Switch XPO releases from raw github to download.

seger:


seger:
@matentzn Here is the updated XPO config.

matentzn (approved):
Looks good to me. See https://github.com/obophenotype/xenopus-phenotype-ontology/issues/156 for an important change on the XPO side.

PR #875, opened 2022-10-11
Update swo.yml

allysonlister:
As a consequence of the update of the repo to build using the ODK.

### DIFF
--- a/config/swo.yml
+++ b/config/swo.yml
@@ -14,3 +14,7 @@ example_terms:
 entries:
 - prefix: /tracker/
   replacement: https://github.com/allysonlister/swo/issues
+
+## generic fall-through, serve direct from github by default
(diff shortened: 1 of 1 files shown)


jamesaoverton:
The first two changes are fine, but are you sure that you want to remove the last `prefix: /` entry? The goal is to keep PURLs working. Is this rule not used for anything?

allysonlister:
Thanks James! Perhaps you can help me a little here. I was comparing the swo file to the new t4fs and thought that perhaps this last bit was unnecessary as the t4fs doesn't have it. I think I was hasty in making the change rather than asking about it. 😉 What does this last bit of the file do?

allysonlister:
Thank you - I have reinstated this - I appreciate the help!

PR #882, opened 2022-11-29
updated the term_browser redirect

sabrinatoro:
as per Melissa's request, I updated the redirect such that http://purl.obolibrary.org/obo/VBO_0011418 would redirect to OLS instead of ontobee.

### DIFF
--- a/config/vbo.yml
+++ b/config/vbo.yml
@@ -3,7 +3,7 @@
 idspace: VBO
 base_url: /obo/vbo
 
-term_browser: ontobee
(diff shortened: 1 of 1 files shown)


jamesaoverton:
I think this change is correct, so I'll merge it.

While I was testing it, I noticed that the example term VBO_0000001 does not seem to exist. Is there another example term that you suggest?

sabrinatoro:
Thank you @jamesaoverton. 
I will update the example. Thank you for catching this issue!

PR #885, opened 2022-12-10
Update dron.yml

hoganwr:
Deleting references to the old repository on bitbucket. Dron is now maintained and hosted on GitHub

### DIFF
--- a/config/dron.yml
+++ b/config/dron.yml
@@ -31,39 +31,6 @@ entries:
 - exact: /dev/dron-ingredient.owl
   replacement: https://s3.amazonaws.com/drugontology/dev/dron-ingredient.owl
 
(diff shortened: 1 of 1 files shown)


jamesaoverton:
PURLs should be permanent. Can these be pointed to GitHub?

hoganwr:
Ah that's a great point James, sorry.  Yes, will redirect them instead.

jamesaoverton:
Several seem to be in the release now, and some might have equivalents in <https://github.com/ufbmi/dron/tree/main/src/ontology/imports>.

If there's no match in GitHub you could delete those PURL entries, or you could leave them pointing to the old BitBucket versions. Both have disadvantages.

PR #886, opened 2022-12-10
Update dron.yml

hoganwr:
Updating PURLs that point at the old bitbucket repo to various places in the new repo.  Admittedly there's some "semantic" mismatches (dev PURLs pointing to latest release, chebi/pro modules pointing to _import.owl in ODK, and hand/upper now pointing to dron-edit.owl), but that's better than breaking the PURLs or pointing to a defunct repo that could go away someday.


jamesaoverton:
Thanks!

PR #892, opened 2023-01-09
Fix fall-through error

shawntanzk:
Sorry, just noticed the link for fall-through was wrong when i was testing, fixed in this PR


jamesaoverton:
Ok. It's good to have tests for `prefix` rules to catch that sort of problem: https://github.com/OBOFoundry/purl.obolibrary.org#tests

PR #897, opened 2023-02-07
Adding a product for gaz-countries

cmungall:
See: https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2108

This PR makes the GAZ countries subset resolvable via its own URL, http://purl.obolibrary.org/obo/gaz/gaz-countries.owl

### DIFF
--- a/config/gaz.yml
+++ b/config/gaz.yml
@@ -7,6 +7,7 @@ products:
 - gaz.owl: http://ontologies.berkeleybop.org/gaz.owl
 - gaz.obo: http://ontologies.berkeleybop.org/gaz.obo
 
+entries:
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Since the filename is not `gaz` this should not be a top-level artifact, so it should not be under `products`. The PURL should be <http://purl.obolibrary.org/obo/gaz/gaz-countries.owl> (or something), with an `entry` like this:

```yaml
- exact: /gaz-countries.owl
  replacement: https://raw.githubusercontent.com/EnvironmentOntology/gaz/master/src/ontology/gaz_countries.owl
```

matentzn:
I think we need this anyways.

PR #916, opened 2023-05-13
Create ispo.yml

HuaRui0420:


jamesaoverton:
The OBO PURL system is only for OBO projects. To join OBO, please follow the instructions here <http://obofoundry.org/faq/how-do-i-register-my-ontology.html>.

PR #928, opened 2023-06-30
Update release versions redirects for CL

anitacaron:
This also updates the redirect for the prefix `/about/` to use OLS instead of Ontobee to be aligned with the term browser.

### DIFF
--- a/config/cl.yml
+++ b/config/cl.yml
@@ -105,7 +105,7 @@ entries:
   replacement: https://cell-ontology.googlecode.com/svn/trunk/src/ontology/diffs/
 
 - prefix: /about/
-  replacement: replacement: https://www.ebi.ac.uk/ols/ontologies/cl/terms?iri=http://purl.obolibrary.org/obo/
+  replacement: https://www.ebi.ac.uk/ols/ontologies/cl/terms?iri=http://purl.obolibrary.org/obo/
   tests:
   - from: /about/CL_0000000
     to: https://www.ebi.ac.uk/ols/ontologies/cl/terms?iri=http://purl.obolibrary.org/obo/CL_0000000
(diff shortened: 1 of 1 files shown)


config/cl.yml
```diff
@@ -68,8 +68,35 @@ entries:
 - prefix: /2011-12-13/
   replacement: https://cell-ontology.googlecode.com/svn/releases/2011-12-13/
 
+- regex: ^/releases/(201.*)/
```
jamesaoverton, line 71: We don't use the `regex` system very often. The examples in the README all end with `$` but I don't think that's what you want here.

I *think* you could get what you want with a series of `prefix` rules.

anitacaron: Also, it's conflicting with the new rule for release.

config/cl.yml
```diff
@@ -68,8 +68,35 @@ entries:
...
+    to: https://raw.githubusercontent.com/obophenotype/cell-ontology/v2015-08-08/cl.obo
+
+- regex: ^/releases/(2020.*)/
+  replacement: https://raw.githubusercontent.com/obophenotype/cell-ontology/v$1
+  tests:
+  - from: /releases/2020-01-06/cl.obo
+    to: https://raw.githubusercontent.com/obophenotype/cell-ontology/v2020-01-06/cl.obo
+
+- regex: ^/releases/(2021.*)/
+  replacement: https://raw.githubusercontent.com/obophenotype/cell-ontology/$1
```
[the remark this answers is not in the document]
anitacaron: fixed, thanks

config/cl.yml
jamesaoverton: This is still failing the tests.

Entries are processed in order from top to bottom, which is why I think that a series of `prefix` entries will do what you want.

anitacaron: Will it work to only change the key `regex` to `prefix`?

anitacaron:
Thank you, James! 🙏🏼

jamesaoverton:
I think this is what you want. It passes all the tests you wrote. Good to merge?

anitacaron:
Yes, thank you!

PR #933, opened 2023-07-29
Try to redirect hash # iris in the PURL system

matentzn:
No Idea if this can work.


jamesaoverton:
I don't think this is possible. The hash part is called the "fragment identifier" and (AFAIK) it is not even sent to the server: https://en.wikipedia.org/wiki/URI_fragment

matentzn:
No, it's indeed not; I just read all around the internet and it seems the part after the hash symbol is not really interpreted by the server..

PR #935, opened 2023-07-31
HANCESTRO release 3.0 with new release artefacts

daniwelter:
New PR as I pushed a typo fix that wasn't picked up by the last PR

### DIFF
--- a/config/hancestro.yml
+++ b/config/hancestro.yml
@@ -13,8 +13,8 @@ example_terms:
 entries:
 - exact: /hancestro.owl
   replacement: https://raw.githubusercontent.com/EBISPOT/hancestro/main/hancestro.owl
-- exact: /hancestro_base.owl
-  replacement: https://raw.githubusercontent.com/EBISPOT/hancestro/main/hancestro_base.owl
-- exact: /hancestro_full.owl
-  replacement: https://raw.githubusercontent.com/EBISPOT/hancestro/main/hancestro_full.owl
+- exact: /hancestro-base.owl
+  replacement: https://raw.githubusercontent.com/EBISPOT/hancestro/main/hancestro-base.owl
+- exact: /hancestro-full.owl
+  replacement: https://raw.githubusercontent.com/EBISPOT/hancestro/main/hancestro-full.owl
   
--- a/config/hancestro.yml
+++ b/config/hancestro.yml
@@ -4,7 +4,7 @@ idspace: HANCESTRO
 base_url: /obo/hancestro
 
 products:
-- hancestro.owl: https://raw.githubusercontent.com/EBISPOT/ancestro/master/hancestro_inferred.owl
+- hancestro.owl: https://raw.githubusercontent.com/EBISPOT/hancestro/main/hancestro.owl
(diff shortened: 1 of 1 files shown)


config/hancestro.yml
```diff
@@ -4,16 +4,17 @@ idspace: HANCESTRO
...
+- hancestro.owl: https://raw.githubusercontent.com/EBISPOT/hancestro/main/hancestro.owl
 
 term_browser: ontobee
 example_terms:
 - HANCESTRO_0004
 
 entries:
 - exact: /hancestro.owl
-  replacement: https://raw.githubusercontent.com/EBISPOT/ancestro/master/hancestro.owl
-- exact: /hancestro_inferred.owl
```
matentzn, line 16: It is against the spirit of OBO PURLs to simply remove them. Can you instead redirect this to https://raw.githubusercontent.com/EBISPOT/hancestro/main/hancestro.owl?

config/hancestro.yml
```diff
@@ -4,16 +4,17 @@ idspace: HANCESTRO
...
-  replacement: https://raw.githubusercontent.com/EBISPOT/ancestro/master/hancestro_inferred.owl
-- exact: /hancestro_bfo.owl
-  replacement: https://raw.githubusercontent.com/EBISPOT/ancestro/master/hancestro_bfo.owl
```
matentzn, line 19: See above - can you redirect this to some file, perhaps an empty one with a comment? Does hancestro_bfo.owl not exist anymore?

config/hancestro.yml
daniwelter: I could absolutely redirect but there's no direct equivalent file, so the resulting file would look different. If that's not a problem, I'll put it back in

config/hancestro.yml
daniwelter: The file hancestro_bfo.owl doesn't exist anymore but the artefact does - it's now called hancestro-full.owl and hancestro.owl

daniwelter:
@matentzn config updated as per your comments

config/hancestro.yml
daniwelter: @matentzn the new `hancestro.owl` will be updated with every release. The old `hancestro_inferred.owl` will never be updated again. There is a legacy version of release 2.8 at `https://github.com/EBISPOT/hancestro/blob/main/retired_artefacts/hancestro_inferred.owl` but this is now obviously out of sync with the ontology content and will diverge more as we add new classes. I'm not sure which is better - pointing the old PURL to a new artefact that looks different or pointing it to the static last version of the old artefact that will diverge more with every new release.

matentzn: I personally think its better to link to the retired artefact, because it makes it at least explicit that the file is no longer maintained. But we are not prescriptive here. How do you want to proceed?

daniwelter: Give me 5 minutes to push an updated version of the config pointing to the old file; I think I prefer that approach. Then the PR is ready to merge.

matentzn (approved):

PR #937, opened 2023-08-23
Update mmo.yml

tutajm:
Update to point to the current (updated weekly) owl and obo files for mmo ontology.


matentzn (approved):
Wonderful thanks!

PR #938, opened 2023-08-23
Update xco.yml

tutajm:
Update to point to the current (updated weekly) owl and obo files for xco ontology.


matentzn (approved):
Great, thank you!

PR #939, opened 2023-08-23
Update pw.yml

tutajm:
Update to point to the current (updated weekly) owl and obo files for pw ontology.


matentzn (approved):
Great, thank you!

PR #940, opened 2023-08-23
Update rs.yml

tutajm:
Update to point to the current (updated weekly) owl and obo files for rs ontology.


matentzn (approved):
fantastic, thanks

PR #944, opened 2023-09-11
Update redirections for Uberon

anitacaron:
- Issue related to the `ext.[owl,obo]` redirection https://github.com/obophenotype/uberon/issues/3031
- Issue related to the `composite-[vertebrate,metazoan].[owl,obo]` redirection https://github.com/obophenotype/uberon/issues/2800#issuecomment-1429830341

### DIFF
--- a/config/uberon.yml
+++ b/config/uberon.yml
@@ -21,16 +21,16 @@ entries:
 - exact: /phenoscape-ext.owl
   replacement: http://phenoscape.svn.sourceforge.net/svnroot/phenoscape/trunk/vocab/edit/phenoscape-ext.owl
 
-- exact: /composite-vertebrate.owl
+- exact: /releases/2021-10-01/composite-vertebrate.owl
   replacement: https://github.com/obophenotype/uberon/releases/download/v2021-10-01/composite-vertebrate.owl
 
-- exact: /composite-vertebrate.obo
+- exact: /releases/2021-10-01/composite-vertebrate.obo
   replacement: https://github.com/obophenotype/uberon/releases/download/v2021-10-01/composite-vertebrate.obo
 
-- exact: /composite-metazoan.owl
+- exact: /releases/2021-10-01/composite-metazoan.owl
   replacement: https://github.com/obophenotype/uberon/releases/download/v2021-10-01/composite-metazoan.owl
 
-- exact: /composite-metazoan.obo
+- exact: /releases/2021-10-01/composite-metazoan.obo
   replacement: https://github.com/obophenotype/uberon/releases/download/v2021-10-01/composite-metazoan.obo
 
 - exact: /ext.owl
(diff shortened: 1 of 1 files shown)


config/uberon.yml
```diff
@@ -21,18 +21,24 @@ entries:
 - exact: /phenoscape-ext.owl
   replacement: http://phenoscape.svn.sourceforge.net/svnroot/phenoscape/trunk/vocab/edit/phenoscape-ext.owl
 
-- exact: /composite-vertebrate.owl
+- exact: /releases/2021-10-01/composite-vertebrate.owl
```
matentzn, line 24: Why is this necessary? If you want `/composite-vertebrate.owl` to be dealt with by the fall-thru, what is the point of hard coding a standard redirect here? Why not just delete the whole goal?

anitacaron: To keep this release version, no?

matentzn: - exact: /releases/2021-10-01/composite-vertebrate.owl
   replacement: https://github.com/obophenotype/uberon/releases/download/v2021-10-01/composite-vertebrate.owl

Redirects the release URL ending with `/releases/2021-10-01/composite-vertebrate.owl` to the release download page of the same version.  The following piece (later on in the makefile, will do the exact same thing, unless I am completely besides me..

```
- prefix: /releases/
  replacement: https://github.com/obophenotype/uberon/releases/download/v
```

config/uberon.yml
```diff
@@ -21,17 +21,11 @@ entries:
 - exact: /phenoscape-ext.owl
   replacement: http://phenoscape.svn.sourceforge.net/svnroot/phenoscape/trunk/vocab/edit/phenoscape-ext.owl
 
-- exact: /composite-vertebrate.owl
-  replacement: https://github.com/obophenotype/uberon/releases/download/v2021-10-01/composite-vertebrate.owl
+- exact: /ext.owl
+  replacement: https://github.com/obophenotype/uberon/releases/latest/download/uberon.owl
```
matentzn, line 25: @jamesaoverton ext.owl is redirected to uberon.owl deliberately - they are now one and the same.

jamesaoverton:
Then shouldn't `composite-[vertebrate,metazoan]` now redirect to `ext`? Or redirect them to older versions? We want to avoid dropping PURLs.

matentzn:
That is what I tried to explain to @anitacaron [here](https://github.com/OBOFoundry/purl.obolibrary.org/pull/944#discussion_r1321504764), but maybe I am wrong here? The PURLs still work exactly the same way as they always did!

jamesaoverton:
I'm not sure that I understand. If `exact: /composite-vertebrate.owl` is now being covered by a `prefix` entry, could `/composite-vertebrate.owl` be added to the tests for that `prefix` entry so I can see that it's not being dropped?

matentzn:
Ok, lets do it right - I fixed it! Thanks for protecting our PURL integrity!

PR #945, opened 2023-09-20
Update ado.yml

BideZ:
We updated the PURL of ADO in the yml file. The latest version v2023-09-20 of ADO has been uploaded.

### DIFF
--- a/config/ado.yml
+++ b/config/ado.yml
@@ -17,8 +17,8 @@ entries:
 - prefix: /releases/
   replacement: https://raw.githubusercontent.com/Fraunhofer-SCAI-Applied-Semantics/ADO/v
   tests:
-  - from: /releases/2022-06-11/ado.owl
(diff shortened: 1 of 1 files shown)


jamesaoverton:
This PR just changes one of the tests to use the latest release. That does no harm, but it is not necessary. Without this change, the PURL system will already redirect <http://purl.obolibrary.org/obo/ado/releases/2023-09-20/ado.owl> to <https://raw.githubusercontent.com/Fraunhofer-SCAI-Applied-Semantics/ADO/v2023-09-20/ado.owl>, which is what I think you want.

jamesaoverton:
In checking this, I noticed a problem with today's ADO release: It looks like the release was made before a large number of changes were committed https://github.com/Fraunhofer-SCAI-Applied-Semantics/ADO/compare/v2023-09-20...main

It might be necessary to make a new GitHub release that includes these changes.

BideZ:
Dear Mr. James A. Overton,

Thanks for the feedback. We have now made a new Github release which included the changes that we have made to ADO.

PR #946, opened 2023-09-28
Make TRANS versionIRI resolvable

allenbaron:
To fix the versionIRI resolvability error of the Pathogen Transmission Ontology on the OBO dashboard.

### DIFF
--- a/config/trans.yml
+++ b/config/trans.yml
@@ -12,3 +12,11 @@ products:
 term_browser: ontobee
 example_terms:
 - TRANS_0000000
+
+- regex: ^/obo/trans/releases/([^\s/]+)/(\S+)$
+  replacement: https://raw.githubusercontent.com/DiseaseOntology/PathogenTransmissionOntology/v$1/src/ontology/$2
+  tests:
+  - from: /releases/2022-10-10/trans.owl
+    to: https://raw.githubusercontent.com/DiseaseOntology/PathogenTransmissionOntology/v2022-10-10/src/ontology/trans.owl
+  - from: /releases/2022-10-10/trans.obo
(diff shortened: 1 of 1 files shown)


config/trans.yml
```diff
@@ -12,3 +12,11 @@ products:
...
 example_terms:
 - TRANS_0000000
+
+- regex: ^/obo/trans/releases/([^\s/]+)/(\S+)$
+  replacement: https://raw.githubusercontent.com/DiseaseOntology/PathogenTransmissionOntology/v$1/src/ontology/$2
+  tests:
+  - from: /releases/2022-10-10/trans.owl
+    to: https://raw.githubusercontent.com/DiseaseOntology/PathogenTransmissionOntology/v2022-10-10/src/ontology/trans.owl
+  - from: /releases/2022-10-10/trans.obo
+    to: https://raw.githubusercontent.com/DiseaseOntology/SymptomOntology/v2022-10-10/src/ontology/trans.obo
```
jamesaoverton, line 22: Should this one be `SymptomOntology`?

allenbaron:
Hahaha.... no. Sorry about that. I was lazy and just copied it from the Symptom Ontology. I guess I missed that one.

config/trans.yml
```diff
@@ -12,3 +12,11 @@ products:
 term_browser: ontobee
 example_terms:
 - TRANS_0000000
+
```
jamesaoverton, line 15: Missing an `entries:` line before the first `regex` entry.

jamesaoverton:
Thanks. I prefer to avoid `regex` because it can be a footgun, but it makes sense given the layout of the GitHub repository in this case, and there are tests to verify that it's working as expected.

PR #953, opened 2023-11-29
Create r74n.yml and update obo.yml

slweeb:
Does this look alright? Thanks.


jamesaoverton:
This PURL system is exclusively for OBO Foundry ontologies <https://obofoundry.org/>. See the instructions for submission here: <https://obofoundry.org/faq/how-do-i-register-my-ontology.html>.

PR #957, opened 2023-12-15
Update DISDRIV purl redirects

allenbaron:
These are fixes to ensure DISDRIV passes the OBO dashboard.
- Updated the ontology IRI redirect: DISDRIV's products will live in the src/ontology directory, for the time-being.
- Added versionIRI redirect
- Added a geneic fallback

A new release dated today was created for DISDRIV with corresponding adjustments to files in the repository to appropriately support these redirects. 

Note: I tried to test this with a vagrant VM to ensure everything was working properly but was unable to get the `make all` or `make build` commands to work. It complained about jsonschema exceptions.

### DIFF
--- a/config/disdriv.yml
+++ b/config/disdriv.yml
@@ -3,11 +3,24 @@
 idspace: DISDRIV
 base_url: /obo/disdriv
 
-
 products:
-- disdriv.owl: https://raw.githubusercontent.com/DiseaseOntology/DiseaseDriversOntology/main/src/ontology/releases/disdriv.owl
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Sorry you had trouble running the tests locally. It looks good to me, and the GitHub Actions tests are passing. I'll merge this now. After about 20 minutes the changes should be live and you can double check.

allenbaron:
It's fine. I just wanted to catch any issues so you wouldn't have to, but I really didn't anticipate any. I'll check them out after it goes live. Thanks for the quick response!

PR #966, opened 2024-01-23
Update gno.yml

edwardsnj:
Patching date-based versions...


jamesaoverton:
I spot-checked a few and they looked right.

PR #968, opened 2024-01-24
Update occo.yml

zhengj2007:
Add version IRI redirection

### DIFF
--- a/config/occo.yml
+++ b/config/occo.yml
@@ -28,3 +28,8 @@ entries:
     
 - prefix: /
   replacement: https://raw.githubusercontent.com/Occupation-Ontology/OccO/master/
+
+### Dated OCCO Releases
+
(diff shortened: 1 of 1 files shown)


zhengj2007:
Hi @jamesaoverton, I found the release tag date is different from the version IRI date in the OWL file. Is it the reason that caused the fail? Thanks!

config/occo.yml
```diff
@@ -28,3 +28,8 @@ entries:
     
 - prefix: /
```
jamesaoverton, line 29: Entries are processed in order. This `prefix: /` entry matches everything, so your new entry is never checked. Please move your new entry before this `prefix: /` entry.

zhengj2007:
@jamesaoverton Thanks for your explanation.

PR #969, opened 2024-01-24
Update vo.yml

zhengj2007:
Add version IRI redirection


jamesaoverton:
This change is fine. I have a suggestion: You already have a dated release entry at line 26, for 2009-11-30. You might want to keep all the dated release entries together.

zhengj2007:
@jamesaoverton Thanks for suggestion. I made some changes. I found the two links don't work anymore. So, removed from the file.

jamesaoverton:
Ok. It would be better to keep PURLs for those old versions, but I guess they are lost? I couldn't see anything good in the older commits of the VO GitHub repository.

PR #973, opened 2024-02-01
Add releases entry for VT

jamesaoverton:
By adding this 'prefix' entry, PURLs with the pattern 'http://purl.obolibrary.org/obo/vt/releases/{PATH}' will redirect to 'https://raw.githubusercontent.com/AnimalGenome/vertebrate-trait-ontology/master/releases/{PATH}' . Specifically 'http://purl.obolibrary.org/obo/vt/releases/2024-01-17/vt.owl' will redirect to 'https://raw.githubusercontent.com/AnimalGenome/vertebrate-trait-ontology/master/releases/2024-01-17/vt.owl'.

### DIFF
--- a/config/vt.yml
+++ b/config/vt.yml
@@ -20,3 +20,9 @@ entries:
   tests:
   - from: /about/VT_0000001
     to: http://www.ontobee.org/browser/rdf.php?o=OBA&iri=http://purl.obolibrary.org/obo/VT_0000001
+
+- prefix: /releases/
+  replacement: https://raw.githubusercontent.com/AnimalGenome/vertebrate-trait-ontology/master/releases/
+  tests:
+  - from: /releases/2024-01-17/vt.owl
+    to: https://raw.githubusercontent.com/AnimalGenome/vertebrate-trait-ontology/master/releases/2024-01-17/vt.owl


zhilianghu:
Thanks and it does seem to be the way that should work but it does not - this (copied from above) http://purl.obolibrary.org/obo/vt/releases/2024-01-17/vt.owl end up with a 404 page.  (The same effect if I just click the above link.)

However this URL http://purl.obolibrary.org/obo/vt.owl resolves to https://raw.githubusercontent.com/AnimalGenome/vertebrate-trait-ontology/master/current_release/vt.owl showing our current OWL file.

Could you take a look at our site https://github.com/AnimalGenome/vertebrate-trait-ontology to see where exactly we should place it? (currently I have "/2024-01-17/vt.owl" in both "current_release" and "releases" directories for experimenting.

jamesaoverton:
Yes, I looked at your repository, and the layout is fine. There are several approaches, and they all work.

I will merge this now. The changes will be live in about 20 minutes. You can check that these are working properly:

- http://purl.obolibrary.org/obo/vt.owl should still work
- http://purl.obolibrary.org/obo/vt/releases/2024-01-17/vt.owl should now work

zhilianghu:
Thank you I see it works now.  However could you please take a look at your Dashboard Report -- there are a couple version related problems on VT: https://dashboard.obofoundry.org/dashboard/vt/dashboard.html

jamesaoverton:
I expect those 2 red Xs to become green checkmarks because of this change, whenever the Dashboard is updated next.

PR #975, opened 2024-02-05
Update releases uri

ddooley:


jamesaoverton:
I *think* this will work for the 2024-02-03 release, because I see a 'foodon.owl' file attached to the release when I look here: https://github.com/FoodOntology/foodon/releases/. I think that it won't work for older releases without attached files. Since the 'foodon.owl' files seem to be stored in the repository, you're probably better off redirecting to 'https://raw.githubusercontent.com/FoodOntology/foodon/v'.

jamesaoverton:
@ddooley Can you please reply to my previous comment.

jamesaoverton:
Please reopen if you still want this.

PR #976, opened 2024-02-05
adjust releases uri

ddooley:


jamesaoverton:
This looks like the same situation as #975, so I want to sort that one out first.

jamesaoverton:
Please reopen if you still want this.

PR #985, opened 2024-04-18
Add purls for the NFDI4Plants ontology

Hannah-Doerpholz:
In our project, NFDI4Plants, we created an ontology that we are using in our tools. We auto-create purls based on the obolibrary purl pattern and would like to make the links direct our users to a specific lookup service. I tried to add the correct configuration for the custom term browser in the obo.yml file already, but please have another look to make sure nothing is affecting other projects. This pull request is linked to the issue #986

### DIFF
--- a/config/dpbo.yml
+++ b/config/dpbo.yml
@@ -1,0 +1,17 @@
+# PURL configuration for http://purl.obolibrary.org/obo/dpbo
+
+idspace: DPBO
+base_url: /obo/dpbo
+
+products:
+- dpbo.owl: https://raw.githubusercontent.com/nfdi4plants/nfdi4plants_ontology/main/DPBO.owl
+- dpbo.obo: https://raw.githubusercontent.com/nfdi4plants/nfdi4plants_ontology/main/dpbo.obo
+
+term_browser: custom
+
+example_terms:
+- DPBO_1000110
+
+entries:
+- exact: /wiki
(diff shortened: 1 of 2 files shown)


jamesaoverton:
The custom term browser config looks good. So does the top of the `dbpo.yml`. I have two comments:

1. Do you need a `/wiki` entry that just points to the `main` branch of the repository? If it's not needed, that entry should be removed.

2. OBO requires that old versions of your ontology continue to be accessible. In the <https://raw.githubusercontent.com/nfdi4plants/nfdi4plants_ontology/main/DPBO.owl> file the version IRI is <http://purl.obolibrary.org/obo/dpbo/init/2020-12-01/dpbo.owl>. That's strange because the date is so old, although the file has been recently updated. The `/init/` part is also unusual. Given the PURL configuration in this PR, that version IRI will not resolve, and the OBO Dashboard check for [Versioning](http://obofoundry.org/principles/fp-004-versioning.html) will fail.

If you need help making the version IRI work properly, just let us know.

jamesaoverton:
Sorry, I'm getting ahead of myself. This PURL service is for Open Bio Ontology (OBO) projects. There is a submission and approval process that you will need to go through first: https://obofoundry.org/faq/how-do-i-register-my-ontology.html Please complete that process, and we can reopen this PR.

PR #987, opened 2024-04-19
Update RO replacement to use terms instead of only property

anitacaron:
Related to https://github.com/oborel/obo-relations/issues/799

### DIFF
--- a/config/obo.yml
+++ b/config/obo.yml
@@ -108,13 +108,16 @@ entries:
 # Term redirects for RO
 # Some RO terms have a local ID starting with an alphanumeric character, e.g RO_HOMnnnnn
 - regex: ^/obo/RO_([A-Z]*\d+)$
-  replacement: https://www.ebi.ac.uk/ols/ontologies/ro/properties?iri=http://purl.obolibrary.org/obo/RO_$1
+  replacement: https://www.ebi.ac.uk/ols/ontologies/ro/terms?iri=http://purl.obolibrary.org/obo/RO_$1
   status: see other
   tests:
   - from: /RO_0002202
-    to: https://www.ebi.ac.uk/ols/ontologies/ro/properties?iri=http://purl.obolibrary.org/obo/RO_0002202
+    to: https://www.ebi.ac.uk/ols/ontologies/ro/terms?iri=http://purl.obolibrary.org/obo/RO_0002202
   - from: /RO_HOM0000001
-    to: https://www.ebi.ac.uk/ols/ontologies/ro/properties?iri=http://purl.obolibrary.org/obo/RO_HOM0000001
+    to: https://www.ebi.ac.uk/ols/ontologies/ro/terms?iri=http://purl.obolibrary.org/obo/RO_HOM0000001
+  - from: /RO_0002533
+    to: https://www.ebi.ac.uk/ols/ontologies/ro/terms?iri=http://purl.obolibrary.org/obo/RO_0002533
+    
  
 # Term redirects for GNO
 # Match digits or GlyTouCan ID


config/obo.yml
```diff
@@ -108,13 +108,16 @@ entries:
 # Term redirects for RO
 # Some RO terms have a local ID starting with an alphanumeric character, e.g RO_HOMnnnnn
 - regex: ^/obo/RO_([A-Z]*\d+)$
-  replacement: https://www.ebi.ac.uk/ols/ontologies/ro/properties?iri=http://purl.obolibrary.org/obo/RO_$1
+  replacement: https://www.ebi.ac.uk/ols/ontologies/ro/terms?iri=http://purl.obolibrary.org/obo/RO_$1
   status: see other
   tests:
   - from: /RO_0002202
-    to: https://www.ebi.ac.uk/ols/ontologies/ro/properties?iri=http://purl.obolibrary.org/obo/RO_0002202
+    to: https://www.ebi.ac.uk/ols/ontologies/ro/terms?iri=http://purl.obolibrary.org/obo/RO_0002202
```
matentzn, line 115: I cannot confirm this works.. @jamesamcl

matentzn (changes requested):

config/obo.yml
anitacaron: True, it works for classes, but not for properties.

anitacaron: I think there isn't a solution to this. Why is it using the OLS link in this case?

anitacaron: In CL, we also have properties and classes, and both are resolving correctly. For example, [has_not_completed](http://purl.obolibrary.org/obo/CL_4030044) property and [cell](http://purl.obolibrary.org/obo/CL_0000000) class.

matentzn: Why does it work for CL? This does not make sense to me immediately..

anitacaron: Yeah, that's my question, why the redirect here is using the ols full url, instead of the ro artefacts at the ro repository?

matentzn: I dont know! Are you saying that we should just delete all the RO redirects from here? Apart from the 5 BFO relations, I think this makes sense. Do you believe it would work then?

anitacaron: There's a ro.yaml with the redirects using the artefact from the repository. So not sure the duplication here.

jamesaoverton: There's something that I'm not understanding. Since OLS4 replaced OLS3, when a PURL config has `term_browser: ols` we use URLs with this pattern:

https://github.com/OBOFoundry/purl.obolibrary.org/blob/44fa4b1d41432aa356b792a6190f513d45514815/tools/translate_yaml.py#L311

anitacaron: So we don't need anymore this redirect, right?

matentzn: What about this line:

https://github.com/OBOFoundry/purl.obolibrary.org/blob/44fa4b1d41432aa356b792a6190f513d45514815/tools/translate_yaml.py#L324

RO has some relationships that have not numeric characters in there?

jamesaoverton: Yes, RO has some non-numeric term IDs (e.g. "HOM"), so `term_browser: ols` will not be enough. But I think the target URL should be something like `https://www.ebi.ac.uk/ols4/ontologies/ro/entities/....` using `entities`.

jamesamcl: Both `entities` and `terms` were supposed to work for this use case in OLS4 but there was a bug in the code. It will be fixed in the next OLS4 release.

matentzn: Thanks @jamesamcl 

@jamesaoverton with the entities endpoint poised to work again soon, is the correct course of action to simply delete the specific redirects from the obo redirect file, and add the general ols redirect to the ro.yml file?

jamesaoverton: I think RO still needs custom redirects to handle non-digit characters, e.g. `HOM`. The default just matches digits: https://github.com/OBOFoundry/purl.obolibrary.org/blob/64bba89f3a2b8b546581b78030891b6ab2aa3b78/tools/translate_yaml.py#L324

matentzn: Ah yes sorry. What would happen if we changed the `RedirectMatch` to 

```
'RedirectMatch seeother "^/obo/%s_([0-9A-Za-z]+)$" "%s"'
```

This would cover all cases, including NCIT and PRO?

jamesaoverton: The default rule matches the OBO ID policy. I don't want to change it.

matentzn: Ok. Am I correct then in assuming that we have the following two options:

1. Enumerate all RO relations/classes that do not follow the OBO ID policy and create manual redirects for these, and switch the ro.yml to use OLS provider by default _or_
1. Not use OLS (e.g. ontobee) for RO?

jamesaoverton: I suggest changing these lines in `obo.yml` from `https://www.ebi.ac.uk/ols/ontologies/ro/properties?iri=http://purl.obolibrary.org/obo/` to `https://www.ebi.ac.uk/ols4/ontologies/ro/entities/http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2F`. This works for me for properties like <https://www.ebi.ac.uk/ols4/ontologies/ro/entities/http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FRO_0002202>. It doesn't work for classes like <https://www.ebi.ac.uk/ols4/ontologies/ro/entities/http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FRO_0002532> at the moment, but neither does <http://purl.obolibrary.org/obo/RO_0002532> with the current PURL configuration, and @jamesamcl says it will be fixed in the next release.

You could also enumerate exceptions for the small number of RO classes, and send them to `https://www.ebi.ac.uk/ols4/ontologies/ro/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252F` instead.

jamesamcl: Both of those OLS links work for me now @jamesaoverton

matentzn: 🎉 

Seems everything works now?? No changes needed?

jamesaoverton: Ok, they are both working for me now.

jamesaoverton: @matentzn <http://purl.obolibrary.org/obo/RO_0002532> doesn't work for me, because the current PURL config (before this PR) sends it to <https://www.ebi.ac.uk/ols/ontologies/ro/properties?iri=http://purl.obolibrary.org/obo/RO_0002532> but it's a class not a property.

matentzn: Yes, I hope this will be solved by this line:

https://github.com/OBOFoundry/purl.obolibrary.org/pull/987/files#diff-ad7cd92bb523412d33831672f39c978bd9cd64cd0162445ba3993dba21d22a11R111

anitacaron: So the PR is ready for review.

matentzn (approved):
LGTM - but only @jamesaoverton should merge in this repo

PR #998, opened 2024-06-11
Update maxo.yml, switching maxo to GitHub release PURLs

matentzn:
We moved MAxO to using GitHub releases instead of raw.github locations for the release files. This is always a painful process, as you have cannot have two conflicting wildcard based rewrite rules in nginx, so all have to be done manually. I added a test for all of them.


iimpulse (approved):

matentzn:
Its fine, you can go ahead!

PR #1001, opened 2024-07-18
Update stato.yml

proccaserra:
fixing extension to meet requirements discussed in: https://github.com/EBISPOT/ols4/issues/704


matentzn (approved):
Thank you @proccaserra

PR #1014, opened 2025-01-22
Update ms.yml

mobiusklein:
This is an attempt to solve https://github.com/HUPO-PSI/psi-ms-CV/issues/259 without rewriting how our OWL files are generated.

It keeps the names the same, but should now resolve versioned pURLs.

I am unable to setup a test VM at the moment, so I couldn't test how the blind `prefix: /` rule would interact with `products`. From running the `translate_yaml.py` script, it didn't seem like `products` produced a rule in the generated `.htaccess` file. If this would introduce a conflict, they can be removed?

### DIFF
--- a/config/ms.yml
+++ b/config/ms.yml
@@ -4,7 +4,8 @@ idspace: MS
 base_url: /obo/ms
 
 products:
-- ms.owl: https://github.com/HUPO-PSI/psi-ms-CV/releases/latest/download/psi-ms.owl
+- ms.owl: https://github.com/HUPO-PSI/psi-ms-CV/releases/latest/download/ms.owl
+- psi-ms.owl: https://github.com/HUPO-PSI/psi-ms-CV/releases/latest/download/psi-ms.owl
 - ms.obo: https://github.com/HUPO-PSI/psi-ms-CV/releases/latest/download/psi-ms.obo
 
 term_browser: ontobee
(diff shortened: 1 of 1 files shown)


config/ms.yml
```diff
@@ -13,20 +14,33 @@ example_terms:
...
+  replacement: https://github.com/HUPO-PSI/psi-ms-CV/releases/latest/download/psi-ms.obo
+
+  
+# support directly resolving versioned IRIs
 - prefix: /
-  replacement: https://raw.githubusercontent.com/HUPO-PSI/psi-ms-CV/master/
+  replacement: https://github.com/HUPO-PSI/psi-ms-CV/releases/download/v
+  tests:  
+    - from: /4.1.186/ms.owl
+      to: https://github.com/HUPO-PSI/psi-ms-CV/releases/download/v4.1.86/ms.owl
```
jamesaoverton, line 42: The tests are failing because "4.1.186" != "4.1.86". I assume that's a typo.

config/ms.yml
```diff
@@ -13,20 +14,33 @@ example_terms:
...
 - prefix: /about/
   replacement: https://psidev.info/groups/controlled-vocabularies
 
+- prefix: /home
+  replacement: https://github.com/HUPO-PSI/psi-ms-CV
+
+# intercept exact matches before version prefix blindly grabs everything
+- exact: /ms.owl
+  replacement: https://github.com/HUPO-PSI/psi-ms-CV/releases/latest/download/ms.owl
+- exact: /psi-ms.owl
```
jamesaoverton, line 29: This is the right way to add "psi-ms.owl", which will resolve to <http://purl.obolibrary.org/obo/ms/psi-ms.owl>.

config/ms.yml
```diff
@@ -4,7 +4,8 @@ idspace: MS
 base_url: /obo/ms
 
 products:
-- ms.owl: https://github.com/HUPO-PSI/psi-ms-CV/releases/latest/download/psi-ms.owl
+- ms.owl: https://github.com/HUPO-PSI/psi-ms-CV/releases/latest/download/ms.owl
+- psi-ms.owl: https://github.com/HUPO-PSI/psi-ms-CV/releases/latest/download/psi-ms.owl
```
jamesaoverton, line 8: These products must all be of the form "id_space.file_extension". The ID_SPACEs are governed by OBO, and the relevant one here is "MS"/"ms". You're not allowed to add a new ID_SPACE here with "psi-ms.owl".

config/ms.yml
```diff
@@ -13,20 +14,33 @@ example_terms:
 - MS_0000000
 
 entries:
-- prefix: /releases/
```
jamesaoverton, line 16: PURLs are supposed to be permanent. You can't just drop all these `/releases/` PURLs.

jamesaoverton: It's safer to put all `prefix` entries at the bottom, after all the `exact` entries.

jamesaoverton:
Hi @mobiusklein. Thanks for the PR -- I know it can be daunting. I made some comments. Maybe look at examples in this repo for other ontologies that you're familiar with. If you get stuck, let me know.

Can you confirm for me in some way that you have authority to make these changes? A note from @germa (the official contact listed here https://obofoundry.org/ontology/ms.html) would be one way to let me know.

mobiusklein:
I made some, if not all the requested changes, and fixed the typo in the tests, but wanted to see if the tests would fail still.

@germa hasn't been involved in the PSI-MS CV's maintenance for a long time. I've been a maintainer for over two years, and was the contact person for two previous issues here: 

- https://github.com/OBOFoundry/purl.obolibrary.org/pull/845
- https://github.com/OBOFoundry/purl.obolibrary.org/pull/843

jamesaoverton:
Thanks. I should have checked the source repo for your name, but I was in a rush.

It would be helpful to update the OBO Foundry metadata for MR with the current contact (Principle 11): https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/ontology/ms.md

PR #1020, opened 2025-02-07
Add VT URI pattern for OBA to obo.yml

matentzn:
fixes #1018 

I am not sure I did this correct at all.. especially the regex pattern. Fingers crossed :)

PR #1021, opened 2025-02-10
Update t4fs.yml

allysonlister:
Attempting to align with https://obofoundry.org/principles/checks/fp_004 by adding a /releases/ prefix.

Am I doing so correctly please? Thanks!

### DIFF
--- a/config/t4fs.yml
+++ b/config/t4fs.yml
@@ -26,4 +26,4 @@ entries:
   replacement: https://github.com/terms4fairskills/FAIRterminology/issues
   
 - prefix: /releases/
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Yes, but it would be even better to have an example. I'll add an example pointing to https://github.com/terms4fairskills/FAIRterminology/releases/tag/v2025-02-09. I'll do that and merge this after #1020.

allysonlister:
Thank you very much!

jamesaoverton:
I think this will work, but please wait at least 15 minutes and then test it.

allysonlister:
Works great, thanks!

PR #1026, opened 2025-03-03
Add ExO redirect to obo.yml

matentzn:
Fixes https://github.com/CTDbase/exposure-ontology/issues/39


matentzn:
This was the wrong approach.

PR #1043, opened 2025-05-16
Make sure CL's IRIs for imported files are resolvable.

gouttegd:
This PR adds special rules in the CL configuration for the three types of IRIs that are used in OWL import declarations within the ontology's edit file (imports, components, and pattern-derived definitions). The corresponding files are not release artefacts, so if we redirect those IRIs using the default generic fall-through (which redirects to the download folder of the latest release) they will end in 404 errors. Instead, we redirect those IRIs to point to the master branch.

See https://github.com/obophenotype/cell-ontology/issues/3119.

### DIFF
--- a/config/cl.yml
+++ b/config/cl.yml
@@ -136,6 +136,20 @@ entries:
 - prefix: /bds/
   replacement: https://github.com/obophenotype/brain_data_standards_ontologies/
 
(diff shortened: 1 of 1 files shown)


jamesaoverton (approved):
Looks good. Ready for me to merge?

PR #1052, opened 2025-07-18
Add PURL redirect for DO translations

allenbaron:
Adds redirects for ontology translation files, currently in Spanish (owl/obo) or international (owl) only.

It was decided that these files should have PURLs formatted like this: `/obo/doid/translations/<file name>`, with versionIRIs that have a similar pattern, and be located in the `src/ontology/releases/translations/` directory of HumanDiseaseOntology GitHub repository. This update is necessary because existing DO PURLs do not redirect to this "releases" directory and  these translation files are only saved in this `releases/translations` location in the repository.


jamesaoverton:
I *think* this will work, but make sure to test it (>20 minutes after I merge).

allenbaron:
I will. Sorry I got that year wrong.

allenbaron:
It works for all the files and across multiple releases. Thanks.

PR #1066, opened 2025-11-20
Redirect NCBITaxon PURLs to official NCBI site

### LINKED ISSUE #1010: [NCBITaxon] Use custom browser term
As voted in the [NCBITaxon repository](https://github.com/obophenotype/ncbitaxon/issues/117), it was decided to change the redirect to NCBITaxon instead of Ontobee.

E.g.
https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=9606

jamesaoverton:
Resolves #1010 from https://github.com/obophenotype/ncbitaxon/issues/117

I found this issue at the bottom of my inbox and decided to finally take care of it.

Note that there are non-numeric IDs for ranks and stuff in NCBITaxon, some using `#` fragments. I think that none of them ever resolved. Most were replaced by TAXRANK terms. If someone wants to do a forensic analysis of this, be my guest.

### DIFF
--- a/config/ncbitaxon.yml
+++ b/config/ncbitaxon.yml
@@ -11,7 +11,7 @@ products:
 - ncbitaxon.obo.gz: https://github.com/obophenotype/ncbitaxon/releases/latest/download/ncbitaxon.obo.gz
 - ncbitaxon.json.gz: https://github.com/obophenotype/ncbitaxon/releases/latest/download/ncbitaxon.json.gz
 
-term_browser: ontobee
+term_browser: custom
 example_terms:
 - NCBITaxon_1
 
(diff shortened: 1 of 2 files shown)


jamesaoverton:
Interesting. I'd never seen those before.

jamesaoverton:
@anitacaron I think that you're saying that we should not use the new pages because they have automatically generated textual definitions, but I don't understand why that's a problem.

jamesaoverton:
I merged this. It changes NCBITaxon redirects from Ontobee to the old NCBI Taxonomy pages. I guess well change to the new NCBI Taxonomy pages at some point -- someone else can take care of that.

PR #1097, opened 2026-04-10
Update iao.yml

zhengj2007:
- Updated iao.owl: link to latest IAO release
- Added version IRI (2026-03-30)

### DIFF
--- a/config/iao.yml
+++ b/config/iao.yml
@@ -10,7 +10,7 @@ base_url: /obo/iao
 base_redirect: https://github.com/information-artifact-ontology/IAO 
 
 products:
-- iao.owl: https://raw.githubusercontent.com/information-artifact-ontology/IAO/v2022-11-07/iao.owl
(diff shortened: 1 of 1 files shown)


config/iao.yml
```diff
@@ -10,7 +10,7 @@ base_url: /obo/iao
 base_redirect: https://github.com/information-artifact-ontology/IAO 
 
 products:
-- iao.owl: https://raw.githubusercontent.com/information-artifact-ontology/IAO/v2022-11-07/iao.owl
+- iao.owl: https://raw.githubusercontent.com/information-artifact-ontology/iao/master/iao.owl
```
jamesaoverton: I don't think that this is a good idea. If someone merges something to `master` this PURL will point to it, even if no GitHub release has been made. I suggest either continuing to use a dated PURL, or use the `latest` GitHub release URL.

If you insist on pointing to `master`, you'll have to ensure that it's only updated on release.

jamesaoverton (dismissed):

PR #1111, opened 2026-06-29
Add EMRO configuration file with product URLs

wdduncan:
@jamesaoverton not sure if I added this correctly for a new ontology

### DIFF
--- a/config/emro.yml
+++ b/config/emro.yml
@@ -1,0 +1,16 @@
+idspace: EMRO
+base_url: /obo/emro/
+
+products:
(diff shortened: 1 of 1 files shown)


jamesaoverton:
@wdduncan I think this is right now. Only the primary "emro" file go in `products`, the rest are entries under your idspace like <http://purl.obolibrary.org/obo/emro/emro-non-classified.owl>.

I'll merge now, but if you see any problems let me know.

wdduncan:
Thanks @jamesaoverton !!!

PR #1115, opened 2026-07-13
Update PURL configuration in pbpko.yml

Crispae:
Updating the ontology versioning in OBO format with dated style.
https://github.com/InSilicoVida-Research-Lab/pbpko


jamesaoverton:
I think this is doing what you want now.

The primary goal of the PURL system is to keep links working over time. Your previous PURL configuration included a link to the 1.3.1 release of PBPKO, which someone out there might be using. I restored that one link, removed unnecessary comments, and fixed the failing test.

PR #1122, opened 2026-08-19
Update REX and FIX ontology redirects

cthoyt:
This PR updates the redirects for FIX and REX to point towards Chris Mungall's rescue repositories


cthoyt:
cc @cmungall @StroemPhi

cmungall (approved):

